PR #3, opened 2019-10-07
add synchronizer and cleanup linting issues

kuanf:


kuanf:
force merge since it is reviewed and test passed

PR #25, opened 2019-10-30
update subscription-release

kuanf:
Signed-off-by: Kuan Feng <fengkuan@gmail.com>


pkg/subscriber/namespace/namespace_subscriiber_test.go
```diff
@@ -33,14 +33,14 @@ import (
 var c client.Client
 
 var id = types.NamespacedName{
-	Name:      "endpoint",
```
jnpacker: Just a note that the file name has a typo.

PR #28, opened 2019-10-31
Deploy Helmrepo referred secert and secret func refactor

ianzhang366:
Make Helmrepo secrets available to managed-cluster via a channel.


kuanf:
@ianzhang366 as we talked, please move the store secret logic to subscription controller so that all subscriber can benefit from it. 

And please add OwnerReference (add not reset) to those secrets and configmaps, and label them so that you can handle the `rename` case.

Please also address the incorrect delete behavior here. 

Thanks~

ianzhang366:
Hi @kuanf 

1. I've moved the referred secret/configmap logic to the subscription controller so that all subscribers can benefit from it.

2. The rename case is handled by using multiple labels on a secret/configmap.

3. The incorrect delete behaviour, I merged the correct code from the following, 
https://github.com/IBM/multicloud-operators-subscription/pull/32

The incorrect deletion is tracked by:
https://github.ibm.com/IBMPrivateCloud/CP4MCM/issues/678

Looking forward to your reply.
Ian Zhang

pkg/controller/subscription/referredconfigmap_handler.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	if err != nil {
+		return nil, err
+	}
+
+	return localRef, nil
+}
+
+func getLabelOfSubscription(subName string) *metav1.LabelSelector {
```
kuanf: if it is internal func and is not member of a struct, let's put it into utils. 

I don't think the name matches what the function does here.

kuanf:
@ianzhang366 in general, please merge the to referencexxxx_handler into 1 reference.go
Then we don't need duplicated create and check functions for each kind.

PR #33, opened 2019-11-01
syncrhonizer wait a cycle before start

kuanf:
Signed-off-by: Kuan Feng <fengkuan@gmail.com>


kuanf:
verifying it on josh's env before merge it.

PR #38, opened 2019-11-05
Rename kubernetes ignore

rokej:
Rename .kubernetesIgnore to `.kubernetesignore` all lower case to follow `.gitignore` and `.helmignore` file name convention.


kuanf:
shall we make it case insensitive?

rokej:
@kuanf  I can make it case insensitive but I would like to stick to .kubernetesignore all lower case to be similar to .gitignore and .helmignore.

PR #44, opened 2019-11-07
Timewindow

ianzhang366:
Addressing the time-window feature.


deploy/crds/app_v1alpha1_subscription_crd.yaml
```diff
@@ -17,11 +17,12 @@ spec:
     listKind: SubscriptionList
     plural: subscriptions
     singular: subscription
-  scope: Namespaced
+  scope: ""
```
kuanf: it should be namespaced!

ianzhang366: it was changed by the `operator-sdk` command, is it that I have some wrong settings in my env, which leads to thise?

PR #48, opened 2019-11-09
Fixownerref4referredresouce

ianzhang366:


pkg/controller/subscription/reference.go
```diff
@@ -32,9 +33,10 @@ import (
 
 var SecretKindStr = "Secret"
 var ConfigMapKindStr = "ConfigMap"
+var SubscriptionGVK = schema.GroupVersionKind{Group: "", Kind: "Subscription", Version: "v1alpha1"}
 
 //SercertReferredMarker is used as a label key to filter out the secert coming from reference
-var SercertReferredMarker = "IsReferredBySub"
+var SercertReferredMarker = "IsReferredBySub-"
 
 type referredObject interface {
```
kuanf: Actually, you can do this to include 2 interface into 1
```
type referredObject interface {
	runtime.Object
	metav1.Object
}
```

kuanf:
why do you always have sooooooooooo many commits?  something wrong with your personal repo?

pkg/controller/subscription/reference.go
ianzhang366: fixed this.

PR #55, opened 2019-11-13
Issue933

ianzhang366:
Making sure when the time-window changes, the namespace will sync up with channel right away.

https://github.ibm.com/IBMPrivateCloud/CP4MCM/issues/933


kuanf:
@ianzhang366 please update the reference to release repo to latest too. thanks

PR #65, opened 2019-11-21
Annotation RollingUpdateMaxUnavailable need to be copied to subscription deployable

xiangjingli:
1. Annotation RollingUpdateMaxUnavailable need to be copied to subscription deployable
2. expired  subscription deployables failed to delete as the UID for the same subscription deployables could change.

### DIFF
--- a/pkg/controller/mcmhub/hub.go
+++ b/pkg/controller/mcmhub/hub.go
@@ -59,19 +59,7 @@ func (r *ReconcileSubscription) doMCMHubReconcile(sub *appv1alpha1.Subscription)
 	}
 
 	if targetDpl != nil {
-		dplAnno := dpl.GetAnnotations()
-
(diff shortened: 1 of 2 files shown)


pkg/controller/mcmhub/hub.go
```diff
@@ -303,7 +341,7 @@ func (r *ReconcileSubscription) clearSubscriptionDpls(sub *appv1alpha1.Subscript
 		// no longer hub, check owner-reference and delete if it is generated.
 		owners := hubdpl.GetOwnerReferences()
 		for _, owner := range owners {
-			if owner.UID == sub.UID {
+			if owner.Name == sub.Name {
```
kuanf: why did you change from UID to Name?  UID should match too, isn't it the case?

PR #124, opened 2020-03-18
move operator-sdk version to v0.16.0

mikeshng:

We are currently referencing a commit because we needed the helm version update fix. Now operator-sdk has a proper release that we can reference.


mikeshng:
Related to https://github.com/open-cluster-management/multicloud-operators-subscription-release/pull/58

PR #125, opened 2020-03-18
Rj release ff

rokej:


rokej:
Reverting previous commits about enabling feature-ff because feature-ff buildharness does not work for a public repo.

PR #134, opened 2020-03-26
Fix subscription deployable collection

rokej:
This is to fix and improve the way hub subscription collects its deployables list.


rokej:
https://github.com/open-cluster-management/backlog/issues/1172

https://github.com/open-cluster-management/backlog/issues/1153

PR #137, opened 2020-03-28
add requeue for hub mode subscription

ianzhang366:
- re-word error message
- re-queue subscription when it's on hub mode and it's reference configmap is not accessiale yet

### DIFF
--- a/pkg/controller/subscription/subscription_controller.go
+++ b/pkg/controller/subscription/subscription_controller.go
@@ -177,11 +177,12 @@ func (r *ReconcileSubscription) Reconcile(request reconcile.Request) (reconcile.
 		reconcileErr = r.doReconcile(instance)
 
 		instance.Status.Phase = appv1alpha1.SubscriptionSubscribed
-		if err != nil {
+		if reconcileErr != nil {
 			instance.Status.Phase = appv1alpha1.SubscriptionFailed
-			instance.Status.Reason = err.Error()
+			instance.Status.Reason = reconcileErr.Error()
 			klog.Errorf("doReconcile got error %v", err)
 
+			return reconcile.Result{Requeue: true}, r.Status().Update(context.TODO(), instance)
 		}
 	} else {
 		// no longer local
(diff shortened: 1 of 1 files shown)


ianzhang366:
TL;DR

1. the trace `Listing and watching *unstructured.Unstructured` showing up in our log is OK. 
Due to the fact the `client-go` is using `klog` at `V(2) or V(3)`

https://github.com/kubernetes/client-go/blob/bec269661e48cb1e5fbb0d037f356ffe9e9978a0/tools/cache/reflector.go#L383


2. `expected type *v1.ConfigMap, but watch event object had type *unstructured.Unstructured`

This one is coming from the miss-use of the typed and dynamic client in our code. More specifically the following code.

https://github.com/open-cluster-management/multicloud-operators-subscription/blob/cd50ff67edd89326b144551f313ac7fbc9d4820e/pkg/controller/subscription/subscription_controller.go#L288-L300

As you can see, there's a `r.Get()` call, where `r` is using the client from `mgr.GetClient()`. 

This is the client provided by `manager`("sigs.k8s.io/controller-runtime/pkg/manager"), dig into the code, we can see the client implements the `Get` in the following way:

```
// Get retrieves an obj for a given object key from the Kubernetes Cluster.
func (d *DelegatingReader) Get(ctx context.Context, key ObjectKey, obj runtime.Object) error {
	_, isUnstructured := obj.(*unstructured.Unstructured)
	if isUnstructured {
		return d.ClientReader.Get(ctx, key, obj)
	}
	return d.CacheReader.Get(ctx, key, obj)
}
```

In other words, the `Get` method in our call will grab resources from the cache instead of calling to API server.

Then the question is, what's the cache within the manager framework? 

The answer is, cache within the manager is `unstructured.Unstructured`, since the manager supposed to be able to list and watch all kinds of resources. Also, the following implementation of the manager's cache, confirms the theory.

```
func createUnstructuredListWatch(gvk schema.GroupVersionKind, ip *specificInformersMap) (*cache.ListWatch, error) {}
```

For fix this issue, we can either 
- change the way we get the configmap, instead of using the typed way, we can use the `unstructured.Unstructured` to get.
Or
- we can use the `hubclient` to get the configmap directly from the hub, since at the subscription set up stage, we actullay need this one from hub to bootstrap the subscription.

PR #146, opened 2020-03-31
Handle multi-resource YAMLs in github subscription

rokej:


rokej:
https://github.com/open-cluster-management/backlog/issues/1247

pkg/subscriber/github/subscriber_item.go
```diff
@@ -950,33 +959,46 @@ func (ghsi *SubscriberItem) sortResources(repoRoot string, resourcePath string)
...
+			t := kubeResource{}
+			err := yaml.Unmarshal(resources[0], &t)
+
+			if err != nil {
+				fmt.Println("Failed to unmarshal YAML file")
+				// Just ignore the YAML
+				return nil
+			}
+
+			if t.APIVersion != "" && t.Kind != "" {
```
ianzhang366: Hi Roke, 

I'm missing some context over here, why do we want to handle these kinds separately?

rokej: The input to our system is a YAML with multiple kube resources. The output of our system that the kube client API we use takes one kube resource object to apply. So within our system, we need to handle this type of YAML, separate all resources and feed one resource at a time to the kube client.

ianzhang366: Yes, I understand that. 

Maybe my question is not clear, I meat to ask, in this `if` block, we are sperate `ghsi.rbacFiles`, `ghsi.crdsAndNamespaceFiles` against`ghsi.otherFiles `. I mean, why do you need this kind separation.

rokej: When there are mixture of kube resource kinds, we need to ensure that certain kinds of resources are applied first. For example, without a namespace, applying resouces into that namespace will fail. Without RBAC resource, applying other resources depending on the RBAC will fail. Line 150 applies these in such order. This ordering might not be complete today. Using kustomize ensures that the resource kinds are applied in correct order.

ianzhang366: Nice, new knowledge gained, thanks.

I was thinking, we will just let the resource fail, and the second or later sync loop will pick it up and apply.

PR #153, opened 2020-04-02
Update .travis.yml

rokej:


rokej:
https://github.com/open-cluster-management/backlog/issues/845

PR #155, opened 2020-04-02
fix spec not revert back to default when its nil

mikeshng:

For https://github.com/open-cluster-management/backlog/issues/993


mikeshng:
merged into release-1.0;0 as well.

PR #157, opened 2020-04-03
Update gitrepo.go

jnpacker:
- Added debug info


jnpacker:
Discard. The problem seems fixed in the latest build of the standalone subscription.

Issue #170, opened 2020-04-21
Appsubs fail to apply CRDs: schemas are required


ahadas:
We've defined an appsub that applies CRD with:
```yaml
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata:
  labels:
    operator.kubevirt.io: ""
  name: kubevirts.kubevirt.io
spec:
  additionalPrinterColumns:
  - JSONPath: .metadata.creationTimestamp
    name: Age
    type: date
  - JSONPath: .status.phase
    name: Phase
    type: string
  group: kubevirt.io
  names:
    categories:
    - all
    kind: KubeVirt
    plural: kubevirts
    shortNames:
    - kv
    - kvs
    singular: kubevirt
  scope: Namespaced
  version: v1alpha3
  versions:
  - name: v1alpha3
    served: true
    storage: true
```

It failed due to:
`E0421 17:05:01.346887       1 synchronizer.go:461] Failed to apply kind template&{map[apiVersion:apiextensions.k8s.io/v1 kind:CustomReso
urceDefinition metadata:map[annotations:map[apps.open-cluster-management.io/hosting-deployable:ec/ec-channel-CustomResourceDefinition-ku
bevirts.kubevirt.io apps.open-cluster-management.io/hosting-subscription:ec/cnv-appsub apps.open-cluster-management.io/sync-source:subgb
k8s-ec/cnv-appsub] labels:map[operator.kubevirt.io:] name:kubevirts.kubevirt.io] spec:map[additionalPrinterColumns:[map[JSONPath:.metada
ta.creationTimestamp name:Age type:date] map[JSONPath:.status.phase name:Phase type:string]] group:kubevirt.io names:map[categories:[all
] kind:KubeVirt plural:kubevirts shortNames:[kv kvs] singular:kubevirt] scope:Namespaced version:v1alpha3 versions:[map[name:v1alpha3 se
rved:true storage:true]]]]}with error:CustomResourceDefinition.apiextensions.k8s.io "kubevirts.kubevirt.io" is invalid: spec.versions[0]
.schema.openAPIV3Schema: Required value: schemas are required`

by changing the versions to:
```yaml
versions:
  - name: v1alpha3
    served: true
    storage: true
    schema:
      openAPIV3Schema:
        type: object
        properties:
          spec:
            type: object
            properties:
              cronSpec:
                type: string
              image:
                type: string
              replicas:
                type: integer
```

[...]

Issue #166, opened 2020-04-20
Entities in (policy.)mcm.ibm.com group are rejected


ahadas:
The following scenario that used to work with previous versions of OCM:
1. Have a local appsub with a github channel
2. Store policies and placementbindings in the repo
3. The policies and placementbindings would be added to the hub cluster.

Doesn't work with the latest version, it fails with [this error](https://github.com/open-cluster-management/multicloud-operators-subscription/blob/7c7d2c9961ff64e89a08803af26cab61532eb6a3/pkg/subscriber/github/subscriber_item.go#L433)

@mdelder @jnpacker  fyi

ahadas:
It doesn't fail on policies and placementbindings anymore but we get the same error on other non-Kubernetes resources:
```
I0421 17:06:26.795155       1 gitrepo.go:369] kubevirt.io/v1alpha3/KubeVirt
W0421 17:06:26.795269       1 gitrepo.go:76] Not a Kubernetes resource
I0421 17:06:26.797963       1 gitrepo.go:386] Multi resource
I0421 17:06:26.798020       1 gitrepo.go:340] otherFiles size ^B
W0421 17:06:26.798052       1 helmrepo.go:373] subsciption.spec.name is missing for subscription: ec/cnv-appsub
W0421 17:06:26.798197       1 gitrepo.go:76] Not a Kubernetes resource
I0421 17:06:26.798479       1 subscriber_item.go:284] Skipping resource
E0421 17:06:26.798635       1 subscriber_item.go:270] Failed to apply a resource, error: Resource kubevirt.io/v1alpha3, Kind=KubeVirt is
 not supported
```

ianzhang366:
When I was trying this  with `apiextensions.k8s.io/v1` on my own, I hit the following part,  #166

```
I0421 17:06:26.795155       1 gitrepo.go:369] kubevirt.io/v1alpha3/KubeVirt
W0421 17:06:26.795269       1 gitrepo.go:76] Not a Kubernetes resource
I0421 17:06:26.797963       1 gitrepo.go:386] Multi resource
I0421 17:06:26.798020       1 gitrepo.go:340] otherFiles size ^B
W0421 17:06:26.798052       1 helmrepo.go:373] subsciption.spec.name is missing for subscription: ec/cnv-appsub
W0421 17:06:26.798197       1 gitrepo.go:76] Not a Kubernetes resource
[...]

PR #175, opened 2020-04-22
Izhang issue 170

ianzhang366:
Addressing:
https://github.com/open-cluster-management/multicloud-operators-subscription/issues/170
and 
https://github.com/open-cluster-management/multicloud-operators-subscription/issues/166

# 170, 
when a user uses a subscription to deploy a CRD at GVK `apiextensions.k8s.io/v1beta1`, our code will convert the `apiextensions.k8s.io/v1`, then the converted template will fail to apply.

User's yaml
```
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata:
  labels:
    operator.kubevirt.io: ""
  name: kubevirts.kubevirt.io
spec:
```

The failure is due to that, for these 2 versions, the k8s have different validation rules. On `v1` there's an extra field required. So simply replace the GVK won't work for us.

At the meantime, if we use `kubectl` to apply the user's sample CRD at `apiextensions.k8s.io/v1beta1`, it will process without any issue. It seems our code is limiting incoming resources.

We are using the following code to limited the resources to be served, mainly limiting resource version.

**pkg/synchronizer/kubernetes/discovery.go**
`resources, err := discovery.NewDiscoveryClientForConfigOrDie(sync.localConfig).ServerPreferredResources()`

The `ServerPreferredResources` would only return the latest resource version to our code, then we convert out-of-date version to the lasted version.

I feel, we should use `ServerResources` to list all the resource versions(GVK) since we should reflect the full capacity of the managed cluster... but maybe there are some other design considerations against doing a full GVK list, can you please comment on this one @xiangjingli  @rokej  @kuanf 

# 166,
It seems when the CRD is updated, the GVK list(cache) is not updated right away, so our code will complain the kind is not supported yet.

I believe we can force the GVK list to rebuild via,

```
pkg/synchronizer/kubernetes/synchronizer.go
sync.rediscoverResource()
```
 The other changes are mainly for the format issues.

### DIFF
--- a/pkg/controller/mcmhub/gitrepo_sync.go
+++ b/pkg/controller/mcmhub/gitrepo_sync.go
@@ -24,11 +24,6 @@ import (
 	"strings"
 
 	"github.com/ghodss/yaml"
-	chnv1 "github.com/open-cluster-management/multicloud-operators-channel/pkg/apis/apps/v1"
-	dplv1 "github.com/open-cluster-management/multicloud-operators-deployable/pkg/apis/apps/v1"
-	appv1 "github.com/open-cluster-management/multicloud-operators-subscription/pkg/apis/apps/v1"
-	appv1alpha1 "github.com/open-cluster-management/multicloud-operators-subscription/pkg/apis/apps/v1"
-	"github.com/open-cluster-management/multicloud-operators-subscription/pkg/utils"
 	"github.com/pkg/errors"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -39,6 +34,12 @@ import (
 	"k8s.io/klog"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/kustomize/pkg/fs"
+
+	chnv1 "github.com/open-cluster-management/multicloud-operators-channel/pkg/apis/apps/v1"
+	dplv1 "github.com/open-cluster-management/multicloud-operators-deployable/pkg/apis/apps/v1"
+	appv1 "github.com/open-cluster-management/multicloud-operators-subscription/pkg/apis/apps/v1"
+	appv1alpha1 "github.com/open-cluster-management/multicloud-operators-subscription/pkg/apis/apps/v1"
+	"github.com/open-cluster-management/multicloud-operators-subscription/pkg/utils"
 )
 
 type kubeResource struct {
--- a/pkg/synchronizer/kubernetes/discovery.go
+++ b/pkg/synchronizer/kubernetes/discovery.go
@@ -67,6 +67,7 @@ var (
 		Group: "apps.open-cluster-management.io",
 		Kind:  "Deployable",
 	}
+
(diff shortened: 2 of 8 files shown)


ianzhang366:
@xiangjingli 

I'm seeing the following test case is failing, 

```
=== RUN   TestGVKValidation
I0422 23:25:44.800863   14121 discovery.go:149] Discovering cluster resources
--- FAIL: TestGVKValidation (0.05s)
    sync_test.go:134: 
        Expected
            <*schema.GroupVersionKind | 0xc0005174a0>: {Group: "apps", Version: "v1beta1", Kind: "StatefulSet"}
        to equal
            <*schema.GroupVersionKind | 0xc0005173b0>: {Group: "apps", Version: "v1", Kind: "StatefulSet"}
```

Do you still remember if we are doing this version change for some special issue or I can just update the test cases so that we are not doing this version change?

ianzhang366:
Hi @rokej @xiangjingli 

I was consulting this with @kuanf. He pointed out that, if we use the `ServerResources` it will add more `watch` on the controller... which could be a very bad idea. 

He suggested, in the long run, we should still use the `ServerPreferredResources` but need to redo some caching logic. So I'm going to hold on to this PR.

For the user's issue, I guess we should direct the user to use `apiextensions.k8s.io/v1` instead, which will skip the version conversion part from our code, meaning we can workaround issue #170 

When I was trying this  with `apiextensions.k8s.io/v1` on my own, I hit the following part,  #166

```
I0421 17:06:26.795155       1 gitrepo.go:369] kubevirt.io/v1alpha3/KubeVirt
W0421 17:06:26.795269       1 gitrepo.go:76] Not a Kubernetes resource
I0421 17:06:26.797963       1 gitrepo.go:386] Multi resource
I0421 17:06:26.798020       1 gitrepo.go:340] otherFiles size ^B
W0421 17:06:26.798052       1 helmrepo.go:373] subsciption.spec.name is missing for subscription: ec/cnv-appsub
W0421 17:06:26.798197       1 gitrepo.go:76] Not a Kubernetes resource
I0421 17:06:26.798479       1 subscriber_item.go:284] Skipping resource
E0421 17:06:26.798635       1 subscriber_item.go:270] Failed to apply a resource, error: Resource kubevirt.io/v1alpha3, Kind=KubeVirt is
 not supported
```
FYI, [yaml used to reproduce](https://github.com/ianzhang366/acm-applifecycle-samples/tree/master/crd-deploy)

Checking the logic, it seems that the CR is not requeue correctly. The flow goes this way,
1. CRD and CR is pulled from github and apply to cluster
2. CRD will be applied, and CR will be rejected, since on manged cluster, we don't have a CRD to support CR
3. CRD applied successfully, and CRD cache is updated
4. GitHub `doSubscription()` kicks in, however, since the git commit ID doesn't change, so we won't retry the CR.

https://github.com/open-cluster-management/multicloud-operators-subscription/blob/4c146908ef91cf8fb3f9da866c1c539f9f5334dd/pkg/subscriber/github/subscriber_item.go#L114-L125

@rokej do you think if we can find a way to re-queue the CR? It seems if we have multiple resources within the same git commit, then the failed one won't be retried.

PR #177, opened 2020-04-23
Fix subscription retry

rokej:


rokej:
https://github.com/open-cluster-management/multicloud-operators-subscription/issues/166

PR #182, opened 2020-04-27
possible synchronizer refactor solution

ianzhang366:
A POC PR for  the following issue:
https://github.com/open-cluster-management/backlog/issues/1948

- Code changes to the following folder, are the one trying to implement the idea, run synchronizer as service, the details of the idea is explained in the above issue,
 `pkg/synchronizer/kubernetes`
Also, with the changes, we don't need the `housekeeping` on synchronizer, since the synchronizer will watch on the sync channel(with buffered size as 4 by default)

- Code changes to the below folder is a small demonstration of the possible changes for the sub-reconcilers code changes, if we use this PR.
`pkg/subscriber/namespace`

Please let me know your thoughts on this refactor. If you guys like it, I will try to apply the new approach to all sub-reconcilers.

FYI, I haven't fixed the test case, but the namespace subscription flow was tested and succeeded with the following examples,

https://github.com/ianzhang366/acm-applifecycle-samples/tree/master/namespace-ops/distribute-resource-via-namespace


pkg/subscriber/github/git_subscriber_item.go
```diff
@@ -323,29 +328,30 @@ func (ghsi *SubscriberItem) subscribeResourceFile(hostkey types.NamespacedName,
...
 		Namespace: dpltosync.Namespace,
 	}
-	kvalid.AddValidResource(*validgvk, hostkey, dplkey)
 
 	pkgMap[dplkey.Name] = true
+	//this is for the adaption of the new synchronizer API
+	klog.V(10).Infof("ingnor this kvalid %v, only exist for adopting synchronizer API", kvalid)
 
-	return nil
+	return err
```
mikeshng, line 354: Double check please. Always return err if it gets this far?

pkg/subscriber/github/git_subscriber_item.go
```diff
@@ -323,29 +328,30 @@ func (ghsi *SubscriberItem) subscribeResourceFile(hostkey types.NamespacedName,
...
 
 	dplkey := types.NamespacedName{
 		Name:      dpltosync.Name,
 		Namespace: dpltosync.Namespace,
 	}
-	kvalid.AddValidResource(*validgvk, hostkey, dplkey)
 
 	pkgMap[dplkey.Name] = true
+	//this is for the adaption of the new synchronizer API
+	klog.V(10).Infof("ingnor this kvalid %v, only exist for adopting synchronizer API", kvalid)
```
mikeshng, line 352: Minor: ingnor should be ignore?

pkg/subscriber/github/git_subscriber_item.go
```diff
@@ -323,29 +328,30 @@ func (ghsi *SubscriberItem) subscribeResourceFile(hostkey types.NamespacedName,
 	pkgMap[dpltosync.GetName()] = true
 
 	klog.V(4).Info("Ready to register template:", hostkey, dpltosync, syncsource)
+	err = ghsi.synchronizer.AddTemplates(syncsource, hostkey, []kubesynchronizer.DplUnit{{Dpl: dpltosync, Gvk: *validgvk}})
 
-	err = ghsi.synchronizer.RegisterTemplate(hostkey, dpltosync, syncsource)
 	if err != nil {
 		err = utils.SetInClusterPackageStatus(&(ghsi.Subscription.Status), dpltosync.GetName(), err, nil)
 
+		ghsi.successful = false
```
mikeshng, line 336: minor: put this before the utils.SetInClusterPackageStatus

pkg/controller/subscription/subscription_controller.go
```diff
@@ -292,8 +294,10 @@ func (r *ReconcileSubscription) doReconcile(instance *appv1alpha1.Subscription)
 			Namespace: instance.Namespace,
 		}
 
-		err = r.hubclient.Get(context.TODO(), subcfgkey, subitem.SubscriptionConfigMap)
-		if err != nil {
+		errLocal := r.Client.Get(context.TODO(), subcfgkey, subitem.SubscriptionConfigMap)
+		errRemote := r.hubclient.Get(context.TODO(), subcfgkey, subitem.SubscriptionConfigMap)
+
+		if errRemote != nil && errLocal != nil {
```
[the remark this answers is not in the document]
ianzhang366: here the idea is checking if the reference `configmap` exists on hub or local. So, if both of them says no, then we really have a problem. Otherwise, we are fine.

pkg/subscriber/github/git_subscriber_item.go
ianzhang366: Hmmm, if we dont have  print statement like this, then we have to the `kvalid` within the function... 

However, with the new code changes, we don't actually need to use the `kvalid`. The idea way is change the function signature to get rid of the `kvalid` parameter... but I don't want to change the original call chain that much... that's why I did the above print..

pkg/subscriber/github/git_subscriber_item.go
ianzhang366: I don't get it... why there's a difference? Please explain a bit more

pkg/subscriber/github/git_subscriber_item.go
ianzhang366: you are right. I should've use `return nil`, which is more friendly. I don't remember why I made this change...

However, using `return err` is equal to `return nil` over here. I mean, if the code gets this far, the only way is the `err == nil`

pkg/controller/subscription/subscription_controller.go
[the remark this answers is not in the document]
ianzhang366: As the `obj := obj`, the `obj` is a pointer within the `for` loop, if we don't do this, then the `obj` pointer might cause us some issue.

The sonar cloud pointed out this one. 

https://github.com/golang/go/issues/20733

pkg/subscriber/github/git_subscriber_item.go
mikeshng: err = xyz()
if err

Should be together for better code readability. 

Also, logic flow wise its better. Setting it to false right away its quick than call the potential longer running function and wait then set.

pkg/subscriber/github/git_subscriber_item.go
mikeshng: I was referring to your spelling mistake.

pkg/subscriber/github/git_subscriber_item.go
mikeshng: Ok you should change it back to return nil in case err is set in the future and code readability.

pkg/subscriber/github/git_subscriber_item.go
ianzhang366: sure

pkg/subscriber/github/git_subscriber_item.go
ianzhang366: sure

pkg/subscriber/github/git_subscriber_item.go
ianzhang366: done

PR #194, opened 2020-05-12
Fix issue 2212

rokej:


rokej:
https://github.com/open-cluster-management/backlog/issues/2212

PR #196, opened 2020-05-13
upgrade operator sdk

mikeshng:

We will need to come back and remove dependencies in

```
// Pinned to kubernetes-1.16.2
replace (
```

once we upgrade `deployable` `channel` repos to use the same `operator-sdk` version as `subscription` and `-release`


rokej:
@mikeshng I think @ianzhang366 added `github.com/golangci/golangci-lint v1.20.0` in go.mod. I am not sure if we need that. Please check with him.

mikeshng:
> @mikeshng I think @ianzhang366 added `github.com/golangci/golangci-lint v1.20.0` in go.mod. I am not sure if we need that. Please check with him.

Ok. It was cleaned up by `go mod tidy` 

@ianzhang366 Do you need that?

Issue #191, opened 2020-05-11
Need to have performance test in place 


ianzhang366:
With the refactor of the synchronizer, we feel that we need a way to tell the performance differences between the new and old synchronizers.

Speaking of the performance, we are talking about the deployment delay between a batch of subscriptions. For instance,

If we apply 100 subscriptions at the same time, then,
a. would the code be able to handle these many subscriptions at once?(would all the subscriptions be processed eventually?)
b. if the first subscription deployed at t=0s, would the last subscription deploy at t=1hr?

In addition, we are assuming 
1. the test is running on 1 hub and 1 managed clusters(meaning the `kubeconfig` of both clusters will be provided) and we are measuring the result on the managed cluster side.
2. each subscription is deployed via namespace channel
3. each subscription is targeted to deploy a `configmap` at the spoke cluster

@xiangjingli 
@rokej 
@mikeshng 
@evelinec 

If you guys have any other ideas please also let me know.

rokej:
@ianzhang366 I like the measurement plan you outlined here. Is the baseline going to be the performance of the old synchronizer? Also, how difficult would it be to increase the concurrency? For example, 10, 20, 30, 50, 100 concurrent subscriptions.

mikeshng:
@ianzhang366 I like the plan but I feel like we should focus on correctness and throughput over performance for now.

mikeshng:
@ianzhang366 test notification

mikeshng:
@ianzhang366 test notification x2

ianzhang366:
@ianzhang366 test notification x2

mikeshng:
@ianzhang366 comment x3

ianzhang366:
@mikeshng @xiangjingli @rokej @jnpacker 

Hi team, here's a quick update for this issue,

# Object
- Build up the benchmark for synchronizer 1.0(batch apply model) and 2.0(the channel version)
- Compare the above 2 models and identify the bottleneck

# Benchmark
1. Test method and measurement 
   - Test method 
   Given hub and spoke cluster `kubeconfig`, then apply N subscriptions on the hub, then measure the resource created time on spoke.
    The subscription is pointing to a namespace channel. The namespace channel has 1 deployable which has a `configmap` as payload.
   
   - Test ENV  
	a. How to run the controller 
	Both tests are running against the binary file of the subscription controller on my Mac.
   
   Sample running command:
   ```go
[...]

PR #202, opened 2020-05-21
fix the retry logic of helmrepo subscriber

ianzhang366:
This fix is making sure the status update error won't override the addToTemplate error.

Without this incorrect overwrite, the helm subscriber logic would retry(after the sync-interval) when the cache channel is full.

Signed-off-by: ianzhang366 <ian.w.zhang@redhat.com>


pkg/subscriber/helmrepo/helm_subscriber_item.go
```diff
@@ -288,27 +288,26 @@ func (hrsi *SubscriberItem) manageHelmCR(indexFile *repo.IndexFile, repoURL stri
 			return err
 		}
 
+		dplkey := types.NamespacedName{
+			Name:      dpl.Name,
+			Namespace: dpl.Namespace,
+		}
+		pkgMap[dplkey.Name] = true
```
mikeshng, line 295: placing this `= true` here is it ok? Or can it be in somewhere more logically accurate.

pkg/subscriber/helmrepo/helm_subscriber_item.go
```diff
@@ -288,27 +288,26 @@ func (hrsi *SubscriberItem) manageHelmCR(indexFile *repo.IndexFile, repoURL stri
...
 
 		if err != nil {
-			klog.Info("eror in registering :", err)
-			err = utils.SetInClusterPackageStatus(&(hrsi.Subscription.Status), dpl.GetName(), err, nil)
+			klog.Error("error in registering :", err)
 
-			if err != nil {
-				klog.Info("error in setting in cluster package status :", err)
+			if serr := utils.SetInClusterPackageStatus(&(hrsi.Subscription.Status), dpl.GetName(), err, nil); serr != nil {
+				klog.Info("error in setting in cluster package status :", serr)
```
mikeshng, line 304: should it be `Info` or `Error` ? Or maybe at leastn `Warn` or something similar

mikeshng:
Thanks Ian. It will be nice to have some test cases that cover this in the other test focus issue you created.

mikeshng:
Besides `helm` I am assuming the other subscribers are ok because there are no changes for them?

pkg/subscriber/helmrepo/helm_subscriber_item.go
ianzhang366: Yup, forgot to change it

pkg/subscriber/helmrepo/helm_subscriber_item.go
ianzhang366: `pkgMap` is more like a placeholder, it not used at all..

Since we don't want to change functions interfaces... I kept the `pkgMap`

ianzhang366:
> Thanks Ian. It will be nice to have some test cases that cover this in the other test focus issue you created.

Yup, we can add more test cases for the helm. But I'm not very familiar with the helm flow in general...

If you can list up the flows you want to be tested, I will add them in.

> Besides `helm` I am assuming the other subscribers are ok because there are no changes for them?

I think we got 
- `github` covered
https://github.com/open-cluster-management/multicloud-operators-subscription/blob/f7297757cdcca97d65ba3ad4c4a529c8c4089524/pkg/subscriber/git/git_subscriber_item.go#L345-L347

- namespace covered
https://github.com/open-cluster-management/multicloud-operators-subscription/blob/f7297757cdcca97d65ba3ad4c4a529c8c4089524/pkg/subscriber/namespace/deployable_reconciler.go#L107-L113

- Objectbucket
I just added similar changes to it.


Maybe we should unify the logic between these subscribers... and unify the update logic between the actual resource and its status.

mikeshng:
> Yup, we can add more test cases for the helm. But I'm not very familiar with the helm flow in general...
If you can list up the flows you want to be tested, I will add them in.

I was thinking is it possible to repeat what Nathan experienced.
Overflow the channel so no helmrelease are generated then once the channel is free up then we see the helmrelease is now properly created. Similar tests can be done for the other types I guess.

The fact that we are seeing channel is full so early its a little bit worrisome as well. Maybe we should have some configurable parameter so we can set the size to higher/lower. What do you think? It might help with testing as well, like we set the buffer to a small size to create the problem then set it high again to see if everything is ok afterwards.

pkg/subscriber/objectbucket/objectbucket_subscriber_item.go
```diff
@@ -216,16 +216,15 @@ func (obsi *SubscriberItem) doSubscription() error {
...
-			if err != nil {
-				klog.Info("error in setting in cluster package status :", err)
+			if serr := utils.SetInClusterPackageStatus(&(obsi.Subscription.Status), dpltosync.GetName(), err, nil); serr != nil {
+				klog.Error("error in setting in cluster package status :", err)
 			}
 
 			pkgMap[dpltosync.GetName()] = true
 
-			continue
+			return err
```
mikeshng, line 227: This is a semi big change behaviour no? Before we would continue doing the others in the loop. 
Is it possible to keep our old behaviour?

ianzhang366: we could but I feel we should let the subscriber fail earlier, and operate in an atomic manner. 

the old logic is trying to do everything at once, if we go a full cache channel on the 5th item, I doubt we will be applied to apply 6th item right away.

ianzhang366:
> > Yup, we can add more test cases for the helm. But I'm not very familiar with the helm flow in general...
> > If you can list up the flows you want to be tested, I will add them in.
> 
> I was thinking is it possible to repeat what Nathan experienced.
> Overflow the channel so no helmrelease are generated then once the channel is free up then we see the helmrelease is now properly created. Similar tests can be done for the other types I guess.
> 
> The fact that we are seeing channel is full so early its a little bit worrisome as well. Maybe we should have some configurable parameter so we can set the size to higher/lower. What do you think? It might help with testing as well, like we set the buffer to a small size to create the problem then set it high again to see if everything is ok afterwards.

I think you are talking about the performance about the cache channel approach. You can check up this post.

https://github.com/open-cluster-management/multicloud-operators-subscription/issues/191

If you want to have all the k8s resource deploy at once, we can add some lock scheme on the cache server-side and fire multiple gorutinue to serve the incoming request. This is given that k8s resources are different kinds. 

If you want to handle the same kind in a huge amount at once, then we need to adjust the `resourcemap` definition.

Again, these depend on what kind of workload we are expecting.

mikeshng:
> > > Yup, we can add more test cases for the helm. But I'm not very familiar with the helm flow in general...
> > > If you can list up the flows you want to be tested, I will add them in.
> > 
> > 
> > I was thinking is it possible to repeat what Nathan experienced.
> > Overflow the channel so no helmrelease are generated then once the channel is free up then we see the helmrelease is now properly created. Similar tests can be done for the other types I guess.
> > The fact that we are seeing channel is full so early its a little bit worrisome as well. Maybe we should have some configurable parameter so we can set the size to higher/lower. What do you think? It might help with testing as well, like we set the buffer to a small size to create the problem then set it high again to see if everything is ok afterwards.
> 
> I think you are talking about the performance about the cache channel approach. You can check up this post.
> 
> #191
> 
> If you want to have all the k8s resource deploy at once, we can add some lock scheme on the cache server-side and fire multiple gorutinue to serve the incoming request. This is given that k8s resources are different kinds.
> 
> If you want to handle the same kind in a huge amount at once, then we need to adjust the `resourcemap` definition.
> 
> Again, these depend on what kind of workload we are expecting.

I was referring to having a configurable channel buffer size. So we can scale up and down easier.

mikeshng:
Retracting my code review for now. More in depth conversation taking place in slack.

PR #208, opened 2020-06-03
Izhang helm anno

ianzhang366:


pkg/subscriber/helmrepo/helm_subscriber_item.go
```diff
@@ -241,10 +249,89 @@ func (hrsi *SubscriberItem) getHelmRepoIndex(client rest.HTTPClient, repoURL str
...
+
+	helms := make([]*releasev1.HelmRelease, 0)
+
+	for pkgName, chartVer := range indexFile.Entries {
+		releaseCRName, err := utils.PkgToReleaseCRName(sub, pkgName)
+		if err != nil {
+			return nil, gerr.Wrapf(err, "failed to generate releaseCRName of helm chart %v for subscription %v", pkgName, sub)
+		}
+
+		helm, err := utils.CreateOrUpdateHelmChart(pkgName, releaseCRName, chartVer, hubClt, chn, sub)
```
xiangjingli, line 325: There may be override part defined in the subscription, so we probably need to call utils.Override to fresh the helmrelease CR. here is the reference code.

https://github.com/open-cluster-management/multicloud-operators-subscription/blob/2159f30ae0e7e77c01b80814c98c1c1c1cda666a/pkg/utils/helmrepo.go#L259

ianzhang366: Good call.

I added the override to the `CreateOrUpdateHelmChart`.

PR #214, opened 2020-06-09
Izhang git helm update v2

ianzhang366:


ianzhang366:
This PR is addressing the issue: 
https://github.com/open-cluster-management/backlog/issues/2632

To do so, we are having, `getGitChart` to get the helm index info from the gitrepo and then reuse the helm parse logic to get the helm resource string defined at:

https://github.com/open-cluster-management/backlog/issues/2405#issuecomment-637077397

In addition, I got some test issue, saying that the k8s cache can't be initialed correctly, so I have to move the code built by @mikeshng  to the `metaupdate.go`  and change the way that how to pass the k8s client and k8s `rest.Config` parameter, to resolve the test error. Hopefully, it is ok to you @mikeshng

mikeshng:
> In addition, I got some test issue, saying that the k8s cache can't be initialed correctly, so I have to move the code built by @mikeshng to the `metaupdate.go` and change the way that how to pass the k8s client and k8s `rest.Config` parameter, to resolve the test error. Hopefully, it is ok to you @mikeshng

I don't mind. Maybe move those functions into its own dedicated file like `helmrelease_util` or something and put a todo or a comment on why they were copied over. Thanks.

ianzhang366:
> > In addition, I got some test issue, saying that the k8s cache can't be initialed correctly, so I have to move the code built by @mikeshng to the `metaupdate.go` and change the way that how to pass the k8s client and k8s `rest.Config` parameter, to resolve the test error. Hopefully, it is ok to you @mikeshng
> 
> I don't mind. Maybe move those functions into its own dedicated file like `helmrelease_util` or something and put a todo or a comment on why they were copied over. Thanks.

Yup, I will put up some comments to explain the reason of this duplication.

PR #220, opened 2020-06-18
Continuously reconcile differences in cluster resources with repo

rokej:
Git repo subscription used to reconcile a resource only when the commit ID of the subscribed repo is changed and the resource has changed in the repo. Once the resource is applied to cluster, any manual changes to the resource are not reconciled until those conditions are met. This PR is to change the behaviour of the subscription so that it continuously reconciles cluster resources to the repository's resources. The repository is the single source of truth.

### DIFF
--- a/pkg/subscriber/git/git_subscriber.go
+++ b/pkg/subscriber/git/git_subscriber.go
@@ -108,9 +108,6 @@ func (ghs *Subscriber) SubscribeItem(subitem *appv1alpha1.SubscriberItem) error 
 
 	ghs.itemmap[itemkey] = ghssubitem
 
-	// Set it back to false so subscription retries until it is set to true
-	ghssubitem.successful = false
-
 	// If the channel has annotation webhookenabled="true", do not poll the repo.
 	// Do subscription only on webhook events.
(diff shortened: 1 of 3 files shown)


ianzhang366:
Hi, @rokej do you mean after the git resource is deployed to the cluster, then the user's modification of the resource will be overridden based on the git.

In other words, if a user wants to change the end result of a git subscription, the only way(or the correct way) should be changed and commit to the git repo. 

Am I understand the flow correctly?

rokej:
@ianzhang366 Yes, that's correct.

ianzhang366:
Hi @rokej , the code look good to me.

Only 1 minor suggestion, I'm seeing you are passing in `commitID`, but within the function, we are not using the `commitID`. I guess you wanted to use the `commitID` to skip unnecessary download, am I right? If that's the case, can we utilize this `commitID` parameter?

`func (ghsi *SubscriberItem) cloneGitRepo() (commitID string, err error) {`

rokej:
@ianzhang366 That's a very good observation. However, we store git-commit in subscription annotation too. Instead of removing it, I added a line of code to log commit ID so enough information is available for troubleshooting.

PR #227, opened 2020-06-25
ensure helmrelease URLs are valid

mikeshng:

https://github.com/open-cluster-management/backlog/issues/2477


pkg/utils/helmrepo.go
```diff
@@ -540,3 +556,9 @@ func DeleteHelmReleaseCRD(runtimeClient client.Client, crdx *clientsetx.Clientse
 		}
 	}
 }
+
+//IsURL return true if string is a valid URL
+func IsURL(str string) bool {
+	u, err := url.Parse(str)
```
xiangjingli, line 562: According to this - https://stackoverflow.com/questions/31480710/validate-url-with-standard-package-in-go
it looks url.ParseRequestURI is betten than url.Parse for validaing URL string.
 ```url.ParseRequestURI returns an error if the URL is not valid, not an absolute url, etc etc. url.Parse returns valid on almost anything```

The rest of the fix LGTM

PR #230, opened 2020-07-06
Improve subscription status

rokej:
When there is no placement in subscription, 

```
status:
  lastUpdateTime: "2020-07-06T18:57:22Z"
  message: Active
  phase: PropagationFailed
  reason: Placement must be specified
```

If it fails to process git subscription in hub, save why it failed in the subscription status message like this.

```
status:
  lastUpdateTime: "2020-07-06T18:58:54Z"
  message: Active
  phase: PropagationFailed
  reason: 'lstat /tmp/demo-subscription/master/wrongresources: no such file or directory'
```


rokej:
https://github.com/open-cluster-management/backlog/issues/3306

PR #232, opened 2020-07-07
Continuous helmrepo reconcile

rokej:
It looks like this problem has been around for a long time. The subscription controller reconciles only if there is some change in the source helm repository (index.yaml). So if you delete a `helmrelease` but there is no change in the helm repository, the deleted helmrelease will not be recreated.


rokej:
https://github.com/open-cluster-management/backlog/issues/3336

PR #234, opened 2020-07-08
don't update subscription if there is no change in git subscription a…

xiangjingli:
…nnotations

https://github.com/open-cluster-management/backlog/issues/3388

### DIFF
--- a/pkg/controller/mcmhub/gitrepo_sync.go
+++ b/pkg/controller/mcmhub/gitrepo_sync.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"path/filepath"
+	"reflect"
 	"strings"
 
 	"github.com/ghodss/yaml"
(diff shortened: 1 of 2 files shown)


pkg/controller/mcmhub/gitrepo_sync.go
```diff
@@ -116,19 +118,79 @@ func (r *ReconcileSubscription) UpdateGitDeployablesAnnotation(sub *appv1.Subscr
...
+// ifUpdateGitSubscriptionAnnotation compare given annoations between the two subscriptions. return true if no the same
+func ifUpdateGitSubscriptionAnnotation(origsub, newsub *appv1.Subscription) bool {
+	origanno := origsub.GetAnnotations()
+	newanno := newsub.GetAnnotations()
+
+	// 1. compare deployables list annoation
+	origdplmap := make(map[string]bool)
+
+	if origanno != nil {
+		dpls := origanno[appv1.AnnotationDeployables]
```
ianzhang366, line 143: @xiangjingli I feel we should check up the AnnotationTopo as well since we are intent to retire AnnotationDeployables in the future.

PR #246, opened 2020-07-16
recreate the helmrelease cr if it's deleted

mikeshng:

https://github.com/open-cluster-management/backlog/issues/3336


ianzhang366:
@mikeshng maybe we can expand the `sync-interval` at the following section, for example, we multiple the default interval by 3. Or we can ask @xiangjingli  to change the default interval

```
 time.Duration(hrsi.syncinterval)*3 *time.Second, hrsi.stopch)
```
https://github.com/open-cluster-management/multicloud-operators-subscription/blob/4f17dbb580aa4030c1ce1218d6b4b8c371554b12/pkg/subscriber/helmrepo/helm_subscriber_item.go#L96

PR #251, opened 2020-07-22
Subscription status update flow

ianzhang366:
Adding a small doc for explaining the status update flow


ianzhang366:
> It looks more like code comments but OK. Thanks for putting this together.

LOL it is, kind of note when I was looking into the status update chain. I feel it could be useful, so want to host somewhere. Not sure if it's here or somewhere else, any other suggestions?

PR #260, opened 2020-08-03
add to watch channel changes

xiangjingli:

https://github.com/open-cluster-management/backlog/issues/4151

The fix has two parts:

1. add to watch for channel changes. If so, all the subscriptions relative to the channel will be reconciled.

2. After calling the func doMCMHubReconcile, the subscription instance has gotten the latest status, but not updated to k8s yet.  So fetching the subscription from k8s at this moment still get back the original subscription status. The latest status will be cleaned up.

The fix to update subscription status firstly, then fetch the latest subscription instance from k8s and check if the topo annotation exists. If not, the subscription will be re-queued again. 

@ianzhang366  This new fix modifies the original fix for addressing object store slow object download and creation.  Please cross check the new fix doesn't break the object store fix in issue #4106.


ianzhang366:
> The fix to update subscription status firstly, then fetch the latest subscription instance from k8s and check if the topo annotation exists. If not, the subscription will be re-queued again.

The fix for [#4106](https://github.com/open-cluster-management/backlog/issues/4106) basically is doing a similar re-queue. I don't see any conflict

PR #247, opened 2020-07-16
Add insecure-skip-verify support to git clone (#241)

ianzhang366:
* Add insecure-skip-verify support to git clone

Co-authored-by: zhwong@cn.ibm.com <zhwong@cn.ibm.com>

Bridging the commit for the following issue:
https://github.com/open-cluster-management/multicloud-operators-subscription/issues/239

PR #271, opened 2020-08-20
Fix concurrent map access error on master

frankwg:
Again, it's not easy to merge https://github.com/open-cluster-management/multicloud-operators-subscription/pull/247 directly.  I provide the fix based on master branch instead.


ianzhang366:
PR is closed, so close this one

PR #273, opened 2020-08-21
Add printer columns for local placement and time window type

KevinFCormier:
open-cluster-management/backlog#4299

@jlpadilla suggested that any columns we have displayed in Search results should match the output from `oc get ... -o wide`, so I'd like to propose addingadditional printer columns for local placement and the time window type.


xiangjingli:
The pr has been merged to a new one https://github.com/open-cluster-management/multicloud-operators-subscription/pull/301

closing  it.

PR #275, opened 2020-08-28
[WIP] Ansible integeration

ianzhang366:
This PR is intended to implement the [design](https://docs.google.com/document/d/16PnosRzhLP816s9Z7vb2EGMShibOfA_TBk0d5gWF5Ts/edit)

https://github.com/open-cluster-management/backlog/issues/4555

This PR will changes:
1, the CRD updates, since we are going to add the prehook/posthook to `SubscriptionSpec` as optional fields.
2, refactor of some utils functions in order to reuse them for git related options(such as download, sort, parse YAMLs from git source)
3, defined an Hookprocessor interface for the pre/post Hook, the Hookprocessor will be used by the `ReconcileSubscription` for handling the pre/posthook logic
4, refactor the constructor of the `ReconcileSubscription` to accommodate the Hookprocessor

Need to be done:
1, update the subscription status to mirror the ansible job status
2, how to avoid reconcile that come from the subscription status change
3, how to tell if the subscription is actually deployed to the managed clusters

If you have any suggestions, please let me know.

Also, the following lib is used to define the ansibleJob spec. In addition, the following lib has a running controller, which can update the ansbilejob status based on certain annotation. You can use it to do test.

https://github.com/open-cluster-management/ansiblejob-go-lib


pkg/controller/mcmhub/hook.go
```diff
@@ -0,0 +1,289 @@
...
+
+func printJobs(jobs []unstructured.Unstructured, logger logr.Logger) {
+	for _, job := range jobs {
+		logger.V(3).Info(fmt.Sprintf("download jobs %#v", job))
+	}
+}
+
+type SuffixFunc func(*subv1.Subscription) string
+
+func suffixFromUUID(subIns *subv1.Subscription) string {
```
xiangjingli, line 159: `subscription.meta.uid` will not change when the subscription is updated. So if the main subscription is updated, it requires to launch a new pre-/post- ansible job with unique name. If reply on uid, the  suffixFromUUID function will generate the same suffix. That will cause the ansible job with same name updated. no new ansible job is created. 

Notice by current ansible job opeator implementation, only when a new ansible job CR is created, a new ansible tower job will be launched.  So from subscription repo,  we should make sure whenever the main subscription is updated, a new ansible job CR should be created.  

May consider to apply `subscription.spec.resourceVersion` for suffix generation. Its value changes when the subscription is updated.

pkg/controller/mcmhub/hook.go
```diff
@@ -0,0 +1,289 @@
...
+			}
+		}
+
+		return types.NamespacedName{Name: t.GetName(), Namespace: t.GetNamespace()}, nil
+	}
+
+	return types.NamespacedName{}, nil
+}
+
+func (a *AnsibleHooks) isUpdateSubscription(subIns *subv1.Subscription) bool {
```
xiangjingli, line 225: The pre-/post- ansible job could update in git repo, the function should allow the new ansibe job to register to the subscription.

ianzhang366:
> DegisterSubscription -> DeregisterSubscription ?

Yup, good catch

pkg/controller/mcmhub/hook.go
```diff
@@ -0,0 +1,289 @@
...
+}
+
+func isJobRunSuccessful(job *ansiblejob.AnsibleJob, logger logr.Logger) bool {
+	curStatus := job.Status.AnsibleJobResult.Status
+	logger.V(3).Info(fmt.Sprintf("job status: %v", curStatus))
+
+	return strings.EqualFold(curStatus, JobCompleted)
+}
+
+func (a *AnsibleHooks) IsSubscriptionCompleted(subKey types.NamespacedName) (bool, error) {
```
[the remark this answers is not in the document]
ianzhang366: will do

hack/test/tower.ansible.com_ansiblejobs_crd.yaml
```diff
@@ -0,0 +1,105 @@
...
+            type: string
+          kind:
+            description: 'Kind is a string value representing the REST resource this object represents. Servers may infer this from the endpoint the client submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+            type: string
+          metadata:
+            type: object
+          spec:
+            description: AnsibleJobSpec defines the desired state of AnsibleJob
+            properties:
+              extra_vars:
```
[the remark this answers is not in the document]
ianzhang366: will do

pkg/controller/mcmhub/hook.go
ianzhang366: will use the `spec.resourceVersion`

hack/test/tower.ansible.com_ansiblejobs_crd.yaml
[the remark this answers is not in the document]
ianzhang366: @xiangjingli @mikeshng 

The `extra_vars` is defined as `type: object` at the above CRD link, is it possible that you can let me know if there's a golang struct definition I can use or, I will have to come up something by myself?

```
              extra_vars:
                type: object
                x-kubernetes-preserve-unknown-fields: true
```

pkg/controller/mcmhub/hook.go
ianzhang366: will add this check later, the check will be based on the commit id of git

mikeshng (commented):
More changes are expected after the conversation we had today on slack.

- There will be a new CRD update to ansiblejob to have the optional `tower_auth_secret_namespace` field.
- Introduce `pre-hooks` and `post-hooks` folder structure expectancy if the users planned to use subscription with pre/post hooks
- Add `spec.AnsibleHookSecret` (this is mainly a hack for the ACM UI) so that it will replace the AnsibleJob CR's `tower_auth_secret` value and `tower_auth_secret_namespace`

mikeshng (approved):
more fixes to come later as agreed by ACM backend team such as:
- modify sub should launch new instances of ansiblejob - most critical should be fix first before the rest
- test cases for the above - most critical
- instead of filtering out just the time fields filter out all managedfields in annotation
- a spelling mistake paser instead of parser
- variable name to be more clear for postHookRunable
- local sub status update for mcmhub - we might need to discuss more

PR #276, opened 2020-08-28
Fix helmchart deployable name to fix upgrade problem

rokej:


rokej:
https://github.com/open-cluster-management/backlog/issues/4755

PR #283, opened 2020-09-17
watch placement rules on hub

mikeshng:

for https://github.com/open-cluster-management/backlog/issues/4228

watch placement rule changes that are related to the subscription


xiangjingli (approved):
This is good for reconciling subscription CRs when its placementrule CR target clusters have changes. 

The next part is to append the new target cluster list to ansibleJob spec.extra_var.targets_list

PR #292, opened 2020-09-22
force register if the reconcile came from a placementrule update

mikeshng:

### DIFF
--- a/pkg/controller/mcmhub/ansiblejob.go
+++ b/pkg/controller/mcmhub/ansiblejob.go
@@ -16,6 +16,7 @@ package mcmhub
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"strings"
 	"sync"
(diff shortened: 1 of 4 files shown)


xiangjingli (approved):

ianzhang366 (approved):

pkg/controller/mcmhub/ansiblejob.go
```diff
@@ -70,13 +73,59 @@ func (jIns *JobInstances) registryJobs(subIns *subv1.Subscription, suffixFunc Su
 
 		nx := ins.DeepCopy()
 		suffix := suffixFunc(subIns)
+
+		if forceRegister {
```
ianzhang366, line 77: I think the naming should be something like this, which can make the `force` job unique

```
		suffix := suffixFunc(subIns)

		if forceRegister {
			plrSuffixFunc := func(a *subv1.Subscription) string {
				return fmt.Sprintf("-%v-%v", subIns.GetGeneration(), placementRuleRv)
			}

			suffix = plrSuffixFunc(subIns)
		}

		nx.SetName(fmt.Sprintf("%s%s", nx.GetName(), suffix))
		}
```

pkg/controller/mcmhub/mcmhub_controller.go
```diff
@@ -628,9 +628,12 @@ func (r *ReconcileSubscription) IsSubscriptionCompleted(subKey types.NamespacedN
 	}
 
 	// need to wait for managed cluster reporting back
+	// fix: When placmentrule doesn't have target cluster decision list, managed clusters status is empty.
+	// In this case, main subscription does nothing, It should be regarded as complete sub,
+	// so that posthook will be able to execute
 	managedStatus := subIns.Status.Statuses
 	if len(managedStatus) == 0 {
-		return false, nil
+		return true, nil
```
ianzhang366, line 636: @xiangjingli 

hmmm, I was thinking post hook is triggered after all the managed cluster is done.
If we allow the empty status, then right after the main reconcile, the post hook will be executed...

to address the new case, we might want to check if the decision list is empty or not, if it’s empty, we can return true right aways, if it’s not, then we will have to wait for the managed cluster to report back

PR #296, opened 2020-09-24
adding the local == false flag

ianzhang366:

### DIFF
--- a/pkg/controller/mcmhub/mcmhub_controller.go
+++ b/pkg/controller/mcmhub/mcmhub_controller.go
@@ -550,7 +550,7 @@ func (r *ReconcileSubscription) Reconcile(request reconcile.Request) (result rec
 				}
 			}
 		}
-	} else if pl.Local != nil && *(pl).Local { //local: true
+	} else { //local: true and handle change true to false
 		// no longer hub subscription
 		err = r.clearSubscriptionDpls(instance)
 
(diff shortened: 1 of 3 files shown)


mikeshng (approved):

pkg/controller/mcmhub/mcmhub_controller.go
```diff
@@ -550,7 +550,7 @@ func (r *ReconcileSubscription) Reconcile(request reconcile.Request) (result rec
 				}
 			}
 		}
-	} else { //local: true
+	} else if *(pl).Local { //local: true
```
mikeshng, line 553: maybe its safer if you do this: `(pl).Local != nil && *(pl).Local`

mikeshng (approved):

pkg/controller/mcmhub/mcmhub_controller.go
xiangjingli: or like this to make it further safer :-)

if pl != nil && pl.Local != nil && *(pl).Local {
...
}

mikeshng: Xiangjing the pl nill check is done here https://github.com/open-cluster-management/multicloud-operators-subscription/blob/master/pkg/controller/mcmhub/mcmhub_controller.go#L527

xiangjingli (approved):

PR #300, opened 2020-09-26
[WIP] make sure per reconcile, only 1 modification is made to k8s

ianzhang366:

### DIFF
--- a/pkg/controller/mcmhub/hook_test.go
+++ b/pkg/controller/mcmhub/hook_test.go
@@ -656,6 +656,14 @@ var _ = Describe("given a subscription pointing to a git path,where both pre and
 				return err
 			}
 
+			if err := ManagedClusterUpdateHubStatus(k8sClt, subKey, subv1.SubscriptionSubscribed); err != nil {
+				return err
+			}
+
+			if err := UpdateHostDeployableStatus(k8sClt, subKey, dplv1.DeployablePropagated); err != nil {
+				return err
+			}
+
 			for _, l := range aList.Items {
 				fmt.Printf("got ansiblejob %v/%v\n", l.GetNamespace(), l.GetName())
 			}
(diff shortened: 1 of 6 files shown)


xiangjingli (approved):

xiangjingli:
@ianzhang366  Here I added two new fixes

1.  appsub `spec.placment.placementRef` namespace could be empty. As a result, when the ref placementrule target cluster list changes, the pre/post hook ansible jobs force Register in relative appsub won't be fired.
```
  placement:
    placementRef:
      kind: PlacementRule
      name: dev-placementrule
```
The fix is to apply appsub namespace as the PlacementRef namespace.

2.  I noticed there is PropagationFailed error message in propagated subscription deployable in target cluster namespace on hub.  The error message is from channel validation in the pre/post hook registry functionality, which is supposed to run only on hub controller.  But it is called in managed cluster for some reason. 

```
status:
  lastUpdateTime: "2020-09-26T23:09:10Z"
  phase: Deployed
  resourceStatus:
    ansiblejobs: {}
    lastUpdateTime: "2020-09-26T23:08:12Z"
    phase: PropagationFailed
    reason: 'failed to register hooks, err: Channel.apps.open-cluster-management.io
      "git" not found'
```
That is weird as we start different controllers when the manger is started in hub and managed clusters. So mcmhub_controller reconcile should not be triggered in managed cluster. But it happened.

For starting the hub cluster controller
https://github.com/open-cluster-management/multicloud-operators-subscription/blob/13940f9767cb652fb942efb05d1efd1f5dc7a114/cmd/manager/exec/manager.go#L122

For starting the managed cluster controller
https://github.com/open-cluster-management/multicloud-operators-subscription/blob/13940f9767cb652fb942efb05d1efd1f5dc7a114/cmd/manager/exec/manager.go#L139

To avoid this happen, the fix is to move the pre/post hook registry implementation under this if block, Then pre/post hook registry is guaranteed  to run only in hub subscription.
```
	if pl == nil {
          ...
	} else if pl != nil && (pl.PlacementRef != nil || pl.Clusters != nil || pl.ClusterSelector != nil) {
          ...
          // pre/post hook registry 
          //  apply prehook
          //  re-queue the main appsub if prehook is not complete 
          ...
      } else {
          ...
      }

I have verified the fixes on UI cluster. Please double verify it on your cluster on Monday when you have cycle.

ianzhang366:
> 1. I noticed there is PropagationFailed error message in propagated subscription deployable in target cluster namespace on hub.  The error message is from channel validation in the pre/post hook registry functionality, which is supposed to run only on hub controller.  But it is called in managed cluster for some reason.

@xiangjingli  for this issue, I feel it could be the case the hub is pointing to itself, since the top level phase is `Deploy`

PR #301, opened 2020-09-27
convert placementrule CRD apiversion to v1beta1 for supporting <4.5 v…

xiangjingli:
…ersions of OCP

The apiVersion of the `kind:CustomResourceDefinition` in all of our CRDs  has to be `v1beta1` for backporting <4.5 versions of OCP.  Or the CRD will fail to be applied to OCP <4.5.  OCP 4.4 is still supported by ACM. 

```
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
```
Please note our appsub CRDs version are still remained as V1 
```
apiVersion: apps.open-cluster-management.io/v1
kind: Subscription

### DIFF
--- a/deploy/crds/apps.open-cluster-management.io_subscriptions.yaml
+++ b/deploy/crds/apps.open-cluster-management.io_subscriptions.yaml
@@ -11,6 +11,12 @@ spec:
   - JSONPath: .metadata.creationTimestamp
     name: Age
     type: date
+  - JSONPath: .spec.placement.local
+    name: Local placement
+    type: boolean
+  - JSONPath: .spec.timewindow.windowtype
+    name: Time window
+    type: string
   conversion:
     strategy: None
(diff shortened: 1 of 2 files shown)


ianzhang366:
@xiangjingli 

since you are modifying the CRD, can you please add the following to the print column

https://github.com/open-cluster-management/multicloud-operators-subscription/pull/273#issuecomment-678447515

xiangjingli:
@ianzhang366  
>since you are modifying the CRD, can you please add the following to the print column
#273 (comment)

Just checked the diff, Some of them have been in the master trunk. Also the CRD can' be generated directly now by controller-gen, as it generates CRDs with V1. We have to convert it to v1beta1 for OCP < 4.5 backport. 

So the new diff only merged the new print columns and the markdown comments in types.go

ianzhang366 (approved):

PR #310, opened 2020-10-02
Create git webhook listener service

rokej:
https://github.com/open-cluster-management/backlog/issues/5110

### DIFF
--- a/pkg/webhook/listener/webhook_listner.go
+++ b/pkg/webhook/listener/webhook_listner.go
@@ -123,7 +123,11 @@ func (listener *WebhookListener) Start(l <-chan struct{}) error {
 }
 
 // CreateWebhookListener creates a WebHook listener instance
-func CreateWebhookListener(config, remoteConfig *rest.Config, scheme *runtime.Scheme, tlsKeyFile, tlsCrtFile string, createService bool) (*WebhookListener, error) {
+func CreateWebhookListener(config,
+	remoteConfig *rest.Config,
+	scheme *runtime.Scheme,
+	tlsKeyFile, tlsCrtFile string,
+	createService bool) (*WebhookListener, error) {
 	if klog.V(utils.QuiteLogLel) {
 		fnName := utils.GetFnName()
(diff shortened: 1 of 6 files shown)


pkg/webhook/listener/webhook_listner.go
```diff
@@ -164,9 +178,112 @@ func CreateWebhookListener(config, remoteConfig *rest.Config, scheme *runtime.Sc
...
 
+func createWebhookListnerService(client client.Client, namespace string) error {
+	var theServiceKey = types.NamespacedName{
+		Name:      serviceName,
+		Namespace: namespace,
+	}
+
+	service := &corev1.Service{}
+
+	if err := client.Get(context.TODO(), theServiceKey, service); err != nil {
```
ianzhang366, line 208: not sure if we need a flow to detect and update the existing service, to make sure the leftover one is also in the desired state.

ianzhang366 (approved):

pkg/webhook/listener/webhook_listner.go
rokej: @ianzhang366 I am not sure what you mean by the leftover one. When ACM is uninstalled, the hub-subscription pod gets deleted which deletes this service due to owner ref.

rokej: @ianzhang366 I am not sure what you mean by the leftover one. When ACM is uninstalled, the hub-subscription pod gets deleted which deletes this service due to owner ref.

ianzhang366: I was thinking the case, that if we patch something to the pod and the service is different then the current logic won't pick the new changes, not sure this is a valid flow you want to look into tho.

rokej: You are right that there could be some change in the service configuration but I do not foresee any change in the near future. If there is change, it should be dealt with at that time. I will go ahead with only necessary code for now. Thanks.

PR #311, opened 2020-10-03
fix managed cluster status update

ianzhang366:

Addressing [issue](https://app.zenhub.com/workspaces/applifecycle-issue-tracker-5e480a1f17e47394a67447b9/issues/open-cluster-management/backlog/5887)

@xiangjingli  this one should fix the status update from the managed cluster to hub.

https://github.com/open-cluster-management/multicloud-operators-subscription/compare/issue-5887?expand=1#diff-54df3d0f5d9d5161cc520b082ec0ec8bL259

^ is the reason why the update failed since the `ResourceStatus` is actually `SubscriptionStatus`, which means the Unmarshal will lose most of the info and leads to false positive of the comparison func.

Not sure if the Unmarshal to `SubscriptionStatus` would cover all cases tho, please advise.

### DIFF
--- a/pkg/controller/mcmhub/hub.go
+++ b/pkg/controller/mcmhub/hub.go
@@ -777,8 +777,8 @@ func (r *ReconcileSubscription) updateSubAnnotations(sub *appv1alpha1.Subscripti
 }
 
 func (r *ReconcileSubscription) updateSubscriptionStatus(sub *appv1alpha1.Subscription, found *dplv1alpha1.Deployable, chn *chnv1alpha1.Channel) error {
-	r.logger.Info("entry doMCMHubReconcile:updateSubscriptionStatus")
-	defer r.logger.Info("exit doMCMHubReconcile:updateSubscriptionStatus")
+	r.logger.Info(fmt.Sprintf("entry doMCMHubReconcile:updateSubscriptionStatus %s", PrintHelper(sub)))
+	defer r.logger.Info(fmt.Sprintf("exit doMCMHubReconcile:updateSubscriptionStatus %s", PrintHelper(sub)))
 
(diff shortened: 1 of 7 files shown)


ianzhang366:
> Signed-off-by: Ian Zhang [izhang@redhat.com](mailto:izhang@redhat.com)
> 
> @xiangjingli this one should fix the status update from the managed cluster to hub.
> 
> https://github.com/open-cluster-management/multicloud-operators-subscription/compare/issue-5887?expand=1#diff-54df3d0f5d9d5161cc520b082ec0ec8bL259
> 
> ^ is the reason why the update failed since the `ResourceStatus` is actually `SubscriptionStatus`, which means the Unmarshal will lose most of the info and leads to false positive of the comparison func.
> 
> Not sure if the Unmarshal to `SubscriptionStatus` would cover all cases tho, please advise.

@xiangjingli after some extra minor fix, now I'm facing a specific edge case, 

if you apply the subscription sample, the status won't update to the hub deployable. Then I restart the controller, then the subscription status will update to the hub deployable.

I feel the issue is over the following logic,

https://github.com/open-cluster-management/multicloud-operators-subscription/blob/4569304e22bc8d6bcd1cc1c9cd6973b2d3dc4e9a/pkg/synchronizer/kubernetes/synchronizer.go#L371-L372

When the deploy of the managed subscription is created without any issue, then the above update logic will not be executed, which leads to the status of the subscription won't update to the host hub deployable.

I'm thinking of adding a status check up logic, between the status in the resource map and the current status from the cluster. if they are not the same, then trigger the update logic.

What do you think about it?

xiangjingli (approved):
LGTM!

PR #313, opened 2020-10-07
fix the git watcher

ianzhang366:

Addressing issue:

https://github.com/open-cluster-management/backlog/issues/5938
https://github.com/open-cluster-management/backlog/issues/5838

Both the above issues are caused by the out-dated commit annotation, the fix will sperate the git commit update to its own logic as `git watcher`. 

The separated logic `git watcher` is running in a `goroutine` which has the same lifecycle as the operator. During the lifecycle, the operator will register/deregister itself to the `git watcher`. Once it's registered, then the `git watcher` will responsible to update the subscription commitID annotation when the target branch is updated.

In this way, for issue 5938, we are mapping the branch to multiple subscriptions, it should decrease the outgoing requests to Github.  For issue 5838, the subscription will tightly link to git commit, since the latest commitID will get updated to the subscription by the `git watcher` as long as the `git watch` detects a new commitID.

Also, the `git watcher` provides some helper functions, for mocking test cases, or gets the latest commitID of a given subscription instance when needed.

**Note: the `git watcher` is running the update check every 90seconds by default**

### DIFF
--- a/pkg/controller/mcmhub/gitrepo_sync.go
+++ b/pkg/controller/mcmhub/gitrepo_sync.go
@@ -79,7 +79,7 @@ func (r *ReconcileSubscription) UpdateGitDeployablesAnnotation(sub *appv1.Subscr
 		}
 
 		//making sure the commit id is coming from the same source
-		getCommitFunc := r.hubGitOps.GetCommitFunc()
+		getCommitFunc := r.hubGitOps.GetCommitFunc(sub)
 		commit, err := getCommitFunc(channel.Spec.Pathname,
 			genBranchString(sub),
 			user,
--- a/pkg/controller/mcmhub/gitrepo_sync.go
+++ b/pkg/controller/mcmhub/gitrepo_sync.go
@@ -78,11 +78,13 @@ func (r *ReconcileSubscription) UpdateGitDeployablesAnnotation(sub *appv1.Subscr
 			return false, err
 		}
 
-		commit, err := utils.CloneGitRepo(channel.Spec.Pathname,
-			utils.GetSubscriptionBranch(sub),
+		//making sure the commit id is coming from the same source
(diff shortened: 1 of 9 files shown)


rokej (approved):
In GitWatch(), for every loop, can we print out the git repo/branch and commit comparison? If something is not in sync, I will look at this log first.

pkg/controller/mcmhub/gitrepo_sync.go
```diff
@@ -71,19 +71,8 @@ func (r *ReconcileSubscription) UpdateGitDeployablesAnnotation(sub *appv1.Subscr
...
-		}
-
-		commit, err := utils.CloneGitRepo(channel.Spec.Pathname,
-			utils.GetSubscriptionBranch(sub),
-			user,
-			pwd,
-			utils.GetLocalGitFolder(channel, sub))
-
+		//making sure the commit id is coming from the same source
+		commit, err := r.hubGitOps.GetLatestCommitID(sub)
```
rokej, line 75: @ianzhang366   In https://github.com/open-cluster-management/multicloud-operators-subscription/pull/313/files#diff-42a6db35407823bbecf54fd8e45c7527R171, the commit ID annotation is updated. That will trigger a reconcile. Then I come here. Wouldn't the comparison of the latest commit ID from hubGitOps to the commit ID annotation be always be the same so the deployables will not be regenerated?

xiangjingli (approved):

pkg/controller/mcmhub/gitrepo_sync.go
ianzhang366: good catch, maybe we need to loosen up this one, to make sure the deployable annotation is re-calculated, what do you say

		if !strings.EqualFold(annotations[appv1.AnnotationGitCommit], commit) || r.isHookUpdate(annotations, subKey) {

PR #318, opened 2020-10-09
[WIP] Fix roke 109

rokej:


pkg/controller/mcmhub/hub_git.go
```diff
@@ -287,7 +294,8 @@ func (h *HubGitOps) RegisterBranch(subIns *subv1.Subscription) {
 			h.logger.Error(err, "failed to get commitID from initialDownload")
 		}
 
-		setCommitID(subIns, commitID)
```
ianzhang366, line 290: if we don't set over here, then the first pre hook generation will be off for a while... we can set it to the fake commit Id tho

PR #320, opened 2020-10-13
do not process on helm subscription if the server is in a unhealthy state

mikeshng:
For https://github.com/open-cluster-management/backlog/issues/6118

- @rokej  @xiangjingli @ianzhang366 this will impact install (and even after install it will impact stability of the continue deployment) please review with extra caution.
- This change is to make sure we don't process helm subscription if there is unexpected errors doing fetching of get helmrelease or get subscription.

Also updated the standalone CRDs


pkg/subscriber/helmrepo/helm_subscriber_item.go
```diff
@@ -127,10 +127,37 @@ func (hrsi *SubscriberItem) doSubscription() {
...
+				klog.Error("Failed to determine if HelmRelease status is populated: ", err)
+
+				hrsi.success = false
+
+				return
+			}
+		}
+
+		if isHashDiff || isUnsuccessful || !existsHelmRelease || !populatedHelmReleaseStatus {
+			klog.Infof("Processing Helm Subscription... isHashDiff=%v isUnsuccessful=%v existsHelmRelease=%v populatedHelmReleaseStatus=%v",
```
rokej, line 158: I like this logging.

mikeshng: Trying to follow our new logging guideline :)

xiangjingli (approved):

ianzhang366 (approved):

PR #322, opened 2020-10-14
fix a bug where deregistering helmrelease is happening unintentionally

mikeshng:

related to https://github.com/open-cluster-management/backlog/issues/6118

### DIFF
--- a/pkg/subscriber/helmrepo/helm_subscriber_item.go
+++ b/pkg/subscriber/helmrepo/helm_subscriber_item.go
@@ -524,9 +524,16 @@ func (hrsi *SubscriberItem) manageHelmCR(indexFile *repo.IndexFile, repoURL stri
 		pkgMap[dplkey.Name] = true
 	}
 
-	if err := dplpro.Units(hrsi.Subscription, hrsi.synchronizer, hostkey, syncsource, pkgMap, dplUnits); err != nil {
-		klog.Warningf("failed to put helm deployables to cache (will retry), err: %v", err)
-		doErr = err
+	if len(dplUnits) > 0 || (len(dplUnits) == 0 && doErr == nil) {
+		if len(dplUnits) == 0 {
+			klog.Warningf("The dplUnits length is 0, this might lead to deregistration for subscription %s/%s",
+				hrsi.Subscription.Namespace, hrsi.Subscription.Name)
+		}
+
+		if err := dplpro.Units(hrsi.Subscription, hrsi.synchronizer, hostkey, syncsource, pkgMap, dplUnits); err != nil {
+			klog.Warningf("failed to put helm deployables to cache (will retry), err: %v", err)
+			doErr = err
+		}
 	}
 
 	return doErr


xiangjingli (approved):

mikeshng:
Hi @ianzhang366 @rokej could you also sign off on this change cus possible impacts install.

pkg/subscriber/helmrepo/helm_subscriber_item.go
```diff
@@ -524,9 +524,11 @@ func (hrsi *SubscriberItem) manageHelmCR(indexFile *repo.IndexFile, repoURL stri
 		pkgMap[dplkey.Name] = true
 	}
 
-	if err := dplpro.Units(hrsi.Subscription, hrsi.synchronizer, hostkey, syncsource, pkgMap, dplUnits); err != nil {
-		klog.Warningf("failed to put helm deployables to cache (will retry), err: %v", err)
-		doErr = err
+	if len(dplUnits) > 0 {
```
ianzhang366, line 527: is that possible the subscription is pointing to from an index with entries to an empty index? 

If so, we also need to cover that flow

rokej: I think this check covers the case where the index is empty because if the index is empty, len(dplUnits) will be 0 so no addTemplate with empty list.

rokej (approved):

pkg/subscriber/helmrepo/helm_subscriber_item.go
mikeshng: I think we should have a more in depth discussion of what it means when entries from the index are removed.

Are users really going to modify the index repo in order to unsubscribe/deregister/uninstall applications?
I think its more likely they just modify/delete the subscription CR itself.

I don't think we should ever delete/deregister apps based on entries from the helm repo index which we just do a simple download and the download might corrupt the file and if we based our delete based on the corrupted file then it might lead to unintended consequences.

But again I think this is more of a discussion in the future. For now I want to do the safe thing which is don't deregister from this flow.

ianzhang366: the case, I'm thinking is, 

if a subscription points to an index with 10 entries, first
then an update of the subscription comes in, then the subscription is pointing to an index with actually 0 entry.

in this case, we are not giving any signal to the synchronizer to remove the old entries

rokej: I see and agree with Ian's point :) In this case, the source of truth is the helm repo and the filtered index file so `if len(dplUnits) > 0 {` check is making several assumptions.

mikeshng: Yep Ian even in your case, I don't think we want to remove old entries. 
Because like I said are users really going to use the index file to remove subscriptions?

I want to avoid the case where we download a corrupted index file and we start removing stuffs unintentionally.

xiangjingli (approved):

ianzhang366 (approved):

PR #329, opened 2020-10-21
skip the target_clusters dupe ansible job check if commitID has been changed

mikeshng:
For https://github.com/open-cluster-management/backlog/issues/6324

It seems like new ansiblejobs are no longer fired if the commit has been changed (but `target_clusters` remain the same)
I think ansiblejobs should always fired if there is a commit and we detected it.

### DIFF
--- a/pkg/controller/mcmhub/ansiblejob.go
+++ b/pkg/controller/mcmhub/ansiblejob.go
@@ -51,7 +51,8 @@ type appliedJobs struct {
 
 func (jIns *JobInstances) registryJobs(gClt GitOps, subIns *subv1.Subscription,
 	suffixFunc SuffixFunc, jobs []ansiblejob.AnsibleJob, kubeclient client.Client,
-	logger logr.Logger, forceRegister bool, placementRuleRv string, hookType string) error {
+	logger logr.Logger, placementDecisionUpdated bool, placementRuleRv string, hookType string,
+	commitIDChanged bool) error {
 	for _, job := range jobs {
 		jobKey := types.NamespacedName{Name: job.GetName(), Namespace: job.GetNamespace()}
 		ins, err := overrideAnsibleInstance(subIns, job, kubeclient, logger, hookType)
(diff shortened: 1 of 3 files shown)


pkg/controller/mcmhub/ansiblejob.go
```diff
@@ -78,14 +79,14 @@ func (jIns *JobInstances) registryJobs(gClt GitOps, subIns *subv1.Subscription,
...
-		if forceRegister && len(jobRecords.Instance) != 0 {
+		if placementDecisionUpdated && len(jobRecords.Instance) != 0 {
 			plrSuffixFunc := func() string {
 				return fmt.Sprintf("-%v-%v", subIns.GetGeneration(), placementRuleRv)
 			}
 
 			suffix = plrSuffixFunc()
 
-			logger.Info("forceRegister suffix is: " + suffix)
+			logger.V(3).Info("placementDecisionUpdated suffix is: " + suffix)
```
ianzhang366, line 89: I dont't think `V(3)` will work for us, if you want it to be noise then `V(InfoLog)`, less noise `V(DebugLog)`

xiangjingli (approved):

pkg/controller/mcmhub/ansiblejob.go
mikeshng: Using `DebugLog` now.

ianzhang366 (approved):

pkg/controller/mcmhub/ansiblejob.go
```diff
@@ -95,7 +96,7 @@ func (jIns *JobInstances) registryJobs(gClt GitOps, subIns *subv1.Subscription,
 		logger.Info(fmt.Sprintf("registered ansiblejob %s", nxKey))
 
 		if _, ok := jobRecords.InstanceSet[nxKey]; !ok {
-			if forceRegister && len(jobRecords.Instance) > 0 {
+			if !commitIDChanged && placementDecisionUpdated && len(jobRecords.Instance) > 0 {
```
rokej, line 99: Can you add some logs here so when this is not run, we know from the logs?

mikeshng: added some more log please see new commit.

rokej (approved):
Thanks Mike

mikeshng:
Undo part of the change in https://github.com/open-cluster-management/multicloud-operators-subscription/pull/328
that related to managed cluster resource status check.

ianzhang366 (approved):

PR #339, opened 2020-11-12
Channel.spec.insecureSkipVerify for helmrelease

rokej:
https://github.com/open-cluster-management/backlog/issues/6860

When a channel connects to a Helm repo server that is secured by a self-signed or custom TLS certificate, we need to reference a config map in a channel to skip the cert verification like below to avoid getting x509: certificate error.

```
apiVersion: apps.open-cluster-management.io/v1
kind: Channel
metadata:
  name: dev-helmrepo
  namespace: dev
spec:
    type: HelmRepo
    pathname: http://kubernetes-charts.storage.googleapis.com/
    configMapRef: 
      name: skip-cert-verify
      apiVersion: v1
      kind: ConfigMap
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: skip-cert-verify
  namespace: dev
data:
  insecureSkipVerify: "true"
```

In 2.2, this is supported but not recommended. The new way is to use `spec.insecureSkipVerify` in the channel.

```
apiVersion: apps.open-cluster-management.io/v1
kind: Channel
metadata:
  name: dev-helmrepo
  namespace: dev
spec:
    type: HelmRepo
    pathname: http://kubernetes-charts.storage.googleapis.com/
    insecureSkipVerify: true
```

### DIFF
--- a/examples/e2e_example/helm_channel_e2e.yaml
+++ b/examples/e2e_example/helm_channel_e2e.yaml
@@ -23,16 +23,6 @@ spec:
     matchLabels:
       name: helm-sub
 ---
-apiVersion: v1
-kind: ConfigMap
-metadata:
-  name: skip-cert-verify
-  namespace: ch-helm
-  labels:
-    name: helm-sub
-data:
-  insecureSkipVerify: "true"
----
 apiVersion: apps.open-cluster-management.io/v1
 kind: Channel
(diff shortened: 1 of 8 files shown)


mikeshng (approved):
/lgtm thanks Roke.

PR #346, opened 2020-11-18
Nested subscriptions should inherit cluster-admin annotation

rokej:
In nested subscriptions, children subscriptions should inherit `cluster-admin:true` annotation from the top subscription because the children subscriptions are created by subscription controllers which is not running as a subscription admin user. 

https://github.com/open-cluster-management/backlog/issues/7171


xiangjingli (approved):
LGTM!

PR #356, opened 2020-11-27
fix make target build-images

mikeshng:


rokej (approved):

xiangjingli (approved):
LGTM. Please  apply the same change to other app lifecycle repos as well when you have cycle.

PR #359, opened 2020-11-30
Try to find kube server api url

rokej:
https://github.com/open-cluster-management/backlog/issues/6793

Try to find the kubernetes API server URL when running on OCP.


xiangjingli (approved):
LGTM

PR #360, opened 2020-11-30
fix hub vs standalone leader id

mikeshng:

for https://github.com/open-cluster-management/backlog/issues/7391


mikeshng:
The `go.mod` changes are result of `go mod tidy`.

I left the changes in there because when the actual build take places, it will run `go mod tidy` as well. So it's better that we have our code to be consistent until we fix the issue of reverting tag.

mikeshng:
Patched the canary build cluster with the fix and it seems to work.

ianzhang366 (approved):

PR #364, opened 2020-12-02
fix message reconcile loop

ianzhang366:

Addressing:
https://github.com/open-cluster-management/backlog/issues/7106

### DIFF
--- a/pkg/utils/subscription.go
+++ b/pkg/utils/subscription.go
@@ -183,12 +183,46 @@ func FilterOutTimeRelatedFields(in *appv1.Subscription) *appv1.Subscription {
 	return out
 }
 
+// assuming the message is only storing the window info, following format:
+// cluster1:active,cluster2:block
+func isSameMessage(aMsg, bMsg string) bool {
+	aMap, bMap := stringToMap(aMsg), stringToMap(bMsg)
+	if len(aMap) != len(bMap) {
+		return false
+	}
+
+	for ak, av := range aMap {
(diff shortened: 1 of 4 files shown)


pkg/utils/subscription.go
```diff
@@ -183,12 +183,46 @@ func FilterOutTimeRelatedFields(in *appv1.Subscription) *appv1.Subscription {
...
+	if len(aMap) != len(bMap) {
+		return false
+	}
+
+	for ak, av := range aMap {
+		if bv, ok := bMap[ak]; !ok || av != bv {
+			return false
+		}
+	}
+
```
xiangjingli, line 199: It is possible that all key/value pairs in the a map exist in the b map, but there are key/value pairs in the b map don't exist in the a map.  In this case, the a Map is not equal to the b map. So the aMap for...loop could be enhanced like this:

	for ak, av := range aMap {
		if bv, ok := bMap[ak]; !ok || av != bv {
			return false
		}
                 delete (bMap, ak)
	}
        
      if len(bMap) > 0  {
                return false 
     }

ianzhang366: we check the len of 2 maps first, if they are not equal, we return `no equal`

then we compare the keys, say if we have 
aMap:
 1: a
 2: b
 3: c

bMap:
 2: b
 3: c
 4: d

In this case, `1` from aMap can't match up from bMap, so not equal. In other
words, we don't need to check from bMap persepctive.

xiangjingli (approved):

PR #400, opened 2021-01-14
Don't add helmrelease controller to the manager when remote subscript…

xiangjingli:
…ion is running on hub cluster

https://github.com/open-cluster-management/backlog/issues/8427
https://github.com/open-cluster-management/backlog/issues/7533

### DIFF
--- a/pkg/utils/subscription.go
+++ b/pkg/utils/subscription.go
@@ -998,8 +998,18 @@ func IsHub(config *rest.Config) bool {
 
 	objlist, err := dl.List(context.TODO(), metav1.ListOptions{})
 	if err != nil {
-		klog.Infof("Listing multiclusterHub resource failed, err: %v", err)
-		return false
+		if kerrors.IsNotFound(err) {
+			klog.Infof("No multiclusterHub resource found, err: %v", err)
+			return false
+		}
+
+		klog.Infof("Listing multiclusterHub resource failed, exit... err: %v", err)
+		os.Exit(1)
+	}
+
+	if objlist == nil {
+		klog.Infof("obj list is nil, exit...")
+		os.Exit(1)
 	}
 
 	objCount := len(objlist.Items)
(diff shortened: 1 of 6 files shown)


ianzhang366 (approved):

pkg/utils/subscription.go
```diff
@@ -978,3 +981,29 @@ func IsSubscriptionBeDeleted(clt client.Client, subKey types.NamespacedName) boo
...
+	}
+
+	dynamicClient := dynamic.NewForConfigOrDie(config)
+
+	dl = dynamicClient.Resource(multiclusterHubGVR)
+
+	objlist, err := dl.List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		klog.Infof("Listing multiclusterHub resource failed, err: %v", err)
+		return false
```
mikeshng, line 1002: Xiangjing I think this return false is dangerous. Because if the apiserver is down for a bit then even if its a hub, it will say its not a hub and we are back to 2 helm controllers again. I think we should error out and let it crash in this case. What do you think?

mikeshng (commented):
Thanks for making this change XJ! Just 1 comment regarding the error handling.

pkg/utils/subscription.go
```diff
@@ -978,3 +981,29 @@ func IsSubscriptionBeDeleted(clt client.Client, subKey types.NamespacedName) boo
...
+	}
+
+	objCount := len(objlist.Items)
```
mikeshng, line 1005: Is it possible to run into a nil pointer here where objlist is nil? Could you please double check?

pkg/utils/subscription.go
mikeshng: I think we can look if there is the mch api/crd exists first if it doesn't exists then its for sure not a hub. But if it does then we proceed with the list which makes the list not fail on errors like no such api/crd. What do you think?

mikeshng (approved):

mikeshng:
XJ not sure how diff is the release 2.1 and release 2.2 branch because the remove operator sdk work. Hopefully its relatively straightforward to pull this change to 2.1.

PR #402, opened 2021-01-15
use option in all the add func

ianzhang366:
Reading our recent changes, it seems we need to do lots of replacing/modify if we change the
`AddFuncs`.

To avoid this(or make the add func signature stable), I put the `CMD option` to its own package.

Then the initialization process would be,
 There's an `options`  variable initialed in the `cmd/exec/options.go`, which is then used by 
`manager.go`. From the ` manager.go`, we just pass along the `options` variable to all the 
`AddFuncs`. At each `AddFuncs`, it will pick what it needs.

In the future, if we are going to change the options for `AddFuncs`, we just need to change the 
option structure and the specific `AddFuncs` instead of all the `AddFuncs` signatures.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -44,7 +44,7 @@ func AddToManager(m manager.Manager, cfg *rest.Config, ops config.SubscriptionCM
 
 	// If remote subscription pod (appmgr) is running in hub, don't add helmrelease controller to the manager,
 	// As there has been a helmrelease controller running in standalone subscription pod
-	if !utils.IsHub(m.GetConfig()) || ops.Standalone {
+	if ops.Standalone || !utils.IsHub(m.GetConfig()) {
 		klog.Info("Add helmrelease controller when the remote subscription is NOT running on hub or standalone subscription")
 
 		for _, f := range AddHelmToManagerFuncs {
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -15,15 +15,15 @@
 package controller
 
 import (
+	"github.com/open-cluster-management/multicloud-operators-subscription/pkg/config"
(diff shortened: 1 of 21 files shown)


xiangjingli (approved):

mikeshng (approved):
/lgtm 
This is an excellent refactoring @ianzhang366 !!! but please fully test with ACM before merge.
Most importantly there should only be one helmrelease controller running even with the local cluster enable.

rokej (approved):
If this does not fix anything, I feel this type of global change should have been made early in the dev cycle. As Mike mentioned, please make sure correct controllers are started in the right environment. Thank you.

ianzhang366:
> If this does not fix anything, I feel this type of global change should have been made early in the dev cycle. As Mike mentioned, please make sure correct controllers are started in the right environment. Thank you.

totally agree, I will merge this one after the release then

pkg/controller/controller.go
```diff
@@ -44,7 +44,7 @@ func AddToManager(m manager.Manager, cfg *rest.Config, syncid *types.NamespacedN
 
 	// If remote subscription pod (appmgr) is running in hub, don't add helmrelease controller to the manager,
 	// As there has been a helmrelease controller running in standalone subscription pod
-	if !utils.IsHub(m.GetConfig()) || standalone {
+	if !utils.IsHub(m.GetConfig()) || ops.Standalone {
```
mikeshng, line 47: @ianzhang366 @xiangjingli do you think we should change this line to
```
if ops.Standalone || !utils.IsHub(m.GetConfig()) {
```
I am not sure if the golang compiler already does this optimization automatically but its better if we just do it ourselves to make sure. I can make the change to this branch if you guys agree.

ianzhang366: good catch, but this PR is planning to be merged after the 2.2 release, if that timeline is ok for you, please make the change

PR #408, opened 2021-01-28
check type before casting value

mikeshng:

Potentially fixes panic error for https://github.com/open-cluster-management/backlog/issues/8928

Need more log from the user to know for sure but this is the only place in the repo that uses `int64` so it should be 99.9% good to go. 

PS. needs to be cherry-pick to 2.0, 2.1 and 2.2 cus its a panic exit.


rokej:
Could `md["replicas"]` being nil cause the problem?

mikeshng:
> Could `md` or `md["replicas"]` being nil cause the problem?
Those should be fine because of the bold check below:

	if md, **ok** := spec.(map[string]interface{}); **ok** {
		if v, **f** := md["replicas"]; **f** {

but having a second look `unstructuredObj` can be nil but it's obviously not the case for this issue but we should double check anyways.

mikeshng:
@rokej fixed now based your comment.

xiangjingli (approved):

PR #410, opened 2021-02-01
3-way merge patch is not working if a spec field such as timeWindow i…

xiangjingli:
…s removed

https://github.com/open-cluster-management/backlog/issues/8975

1.  when app time window is set to "always active", the `spec.timeWindow` is field is removed.  The deletion patch is not added.
2.  patch service resource failed due to immutable field  `clusterIP`

### DIFF
--- a/pkg/synchronizer/kubernetes/synchronizer.go
+++ b/pkg/synchronizer/kubernetes/synchronizer.go
@@ -292,14 +292,6 @@ func (sync *KubeSynchronizer) updateResourceByTemplateUnit(ri dynamic.ResourceIn
 	if merge || isService {
 		if isService {
 			klog.Info("merging services or service account resource")
-
-			if strings.EqualFold(tplunit.GetKind(), "service") &&
-				strings.EqualFold(tplunit.GetAPIVersion(), "v1") {
-				// delete original spec.clusterIP from the service object before merge patch as clusterIP is auto-generated and immutable
-				unstructured.RemoveNestedField(obj.Object, "spec", "clusterIP")
-
-				klog.Infof("new service obj: %#v", obj.Object)
-			}
 		}
 
 		var objb, tplb, pb []byte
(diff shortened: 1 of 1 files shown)


rokej:
@xiangjingli Wouldn't it be better if it always `replaces` subscription CRs it owns? It seems that you had to hack with service clusterIP to get this new merge work. If you still chooses to do the merge this way, please ensure if works with Git resource merge cases.

mikeshng:
For the 3 way json merge patch part, I had to lookup what this means here https://pkg.go.dev/k8s.io/apimachinery/pkg/util/jsonmergepatch

```
CreateThreeWayJSONMergePatch(original, modified, current
```
it **used** to be
```
CreateThreeWayJSONMergePatch(tplb, tplb, objb)
```
which means the part where it "Calculate deletion patch between original and modified" will always produce no delta because it **used** to be `tplb` minus `tplb`

with XJ's fix its now makes more sense:
```
jsonpatch.CreateThreeWayJSONMergePatch(objb, tplb, objb)
```
- "addition-and-change patch between current and modified." which is comparing `objb` and `tplb`
- "deletion patch between original and modified." which is comparing `objb` and `tplb`

Based on the study above I agree with the merge patch change. Please see Roke's comment for the service part. Thanks XJ!

rokej:
@mikeshng @xiangjingli We haven't done any deletion in merge update that is why it was (tplb, tplb, objb) as far as I know.

rokej (changes requested):
Please double-check the outcome of merge reconcile with Git resources.

mikeshng:
@rokej I see. If we never did delete before and this change might introduce delete then it's a pretty big change of behaviour.

xiangjingli:
@rokej  @mikeshng  please review the new fix based on our discussion - Always apply replace to appsub kind resource

mikeshng (approved):

mikeshng (approved):

rokej (approved):

PR #413, opened 2021-02-04
Fix duplicate pre post ansiblejobs

rokej:
https://github.com/open-cluster-management/backlog/issues/8854

Fix creating duplicate prehook and posthook ansible job CRs.


xiangjingli (approved):
LGTM! I feel it is more reliable now :-)

PR #416, opened 2021-02-10
The app sub should be unsubscribed when the app sub time window is bl…

xiangjingli:
…ocked

https://github.com/open-cluster-management/backlog/issues/9310

### DIFF
--- a/pkg/controller/subscription/subscription_controller.go
+++ b/pkg/controller/subscription/subscription_controller.go
@@ -378,23 +378,6 @@ func (r *ReconcileSubscription) doReconcile(instance *appv1.Subscription) error 
 	}
 
 	if sub, ok := r.subscribers[subtype]; ok {
-		// If the app sub time window is blocked, un-subscribe the app sub
-		tw := subitem.Subscription.Spec.TimeWindow
-		if tw != nil {
-			nextRun := utils.NextStartPoint(tw, time.Now())
-			if nextRun > time.Duration(0) {
-				klog.Infof("Subscription %v/%v is unsubscribed as it is currently blocked by the time window, It will be deployed after %v",
-					subitem.Subscription.GetNamespace(),
-					subitem.Subscription.GetName(), nextRun)
-
-				if err := sub.UnsubscribeItem(types.NamespacedName{Name: subitem.Subscription.Name, Namespace: subitem.Subscription.Namespace}); err != nil {
-					klog.Errorf("failed to unsubscribe, error: %+v", err)
-				}
-
-				return nil
-			}
-		}
-
 		if err := sub.SubscribeItem(subitem); err != nil {
 			klog.Errorf("failed to subscribe with subscriber %v, error %+v", subtype, err)
 		}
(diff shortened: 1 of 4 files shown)


pkg/controller/subscription/subscription_controller.go
```diff
@@ -369,6 +378,23 @@ func (r *ReconcileSubscription) doReconcile(instance *appv1.Subscription) error
 	}
 
 	if sub, ok := r.subscribers[subtype]; ok {
+		// If the app sub time window is blocked, un-subscribe the app sub
```
ianzhang366, line 381: not sure if the unsubscribe would delete the deployed resource tho... if so, is it what we want? 

I mean change from active to block, to me, means new objects won't be deployed, and old objects wouldn't be deleted...

again, if unsubscribe won't delete the old resource, then there's not a problem with this  logic

pkg/synchronizer/kubernetes/sync_client.go
```diff
@@ -122,7 +122,7 @@ func (sync *KubeSynchronizer) AddTemplates(subType string, hostSub types.Namespa
 
 	select {
 	case sync.tplCh <- rsOrder:
-		klog.V(1).Info("wrote resource request/order to cache")
+		klog.V(1).Infof("wrote resource request/order to cache: %#v", rsOrder)
```
ianzhang366, line 125: A minor comment,

`rsOrder` might be a giant object... print the whole object out could be too much, maybe we should add an `String()` func to `rsOrder`.

In this way, we can do
`klog.V(1).Infof("wrote resource request/order to cache: %s", rsOrder)`

Then within the string func, we can decide what would be print out for debugging or tracing

rokej:
@xiangjingli If you unsubscribe, it will delete the application. Block means no update to the application but the application should be running. I believe this is what Git subscription does today. It looks like the bug was opened for Helm application.

rokej (changes requested):
@xiangjingli If you unsubscribe, it will delete the application. Block means no update to the application but the application should be running. I believe this is what Git subscription does today. It looks like the bug was opened for Helm application.

pkg/synchronizer/kubernetes/sync_client.go
xiangjingli: yes, let me revert this change

xiangjingli:
@ianzhang366  @rokej   The issue does happen in helm subscription.  I confirmed that in current git subscription, when time window is from active to block, the deployed resources are removed.  So the fix is to follow the git behaviour to all types of subscriptions.

I feel both you and  @thuyn-581  comments  make sense to me   :-)

So Block could mean
1.  no update to the application but the application should be running.  
or 2.   the deployed application should be removed. In next active time window, the application will be deployed again. 

@thuyn-581 , Could you comment which one you prefer to do?  Since the current git subscription does remove the deployed resources when the time window is blocked, it looks 2 is the way to go.  

Besides 1 is a bit surprise to me as well -  When I block the time window, I do expect no resources are deployed. If not, it looks the time window is not working.

pkg/controller/subscription/subscription_controller.go
xiangjingli: As discussed,  when time window is blocked, any new update of the application. will be blocked. But the deployed resources should be remained.

PR #419, opened 2021-02-16
fix community example

mikeshng:


xiangjingli (approved):
LGTM Just double checked the `--debug=true` parameter is supported in channel repo, but not in subscription repo :-)

PR #423, opened 2021-02-18
Seed userid

jnpacker:
* Currently we do not have a way to pass userID and userGroup from a parent subscription to its children.
* This enables that capability for all subscription types. Regardless of whether they are admin or not, so we know where they came from.


rokej (approved):
Thanks Josh

jnpacker:
This will be obsoleted by @philipwu08 work.

PR #431, opened 2021-03-09
possible fix prow unit test

mikeshng:

Unlike the other 5 app repos, this one has `test` folder already in the root directory. I am suspecting this is what's causing prow to hung when running the prow target unit test.


mikeshng:
this only impacts prow. merging for now.

PR #435, opened 2021-03-10
Fix panic when there is no packageOverrides

rokej:
https://github.com/open-cluster-management/backlog/issues/10340


mikeshng (approved):
/lgtm

PR #443, opened 2021-03-11
revert back to v1 crd

mikeshng:

I've fixed the prow test to use a more recent kubebuilder so we should revert these crds back to `v1`.


mikeshng:
This is not needed.

PR #447, opened 2021-03-15
Reconcile rates

rokej:
https://github.com/open-cluster-management/backlog/issues/10101

Currently, subscribed application resources from a Git repo are reconciled every minute. Performance tests show and also customer CaffeineTV noticed that the performance of reconciling existing resources and deploying new resources is degradated as number of subscriptions increases.

Low - unconditional reconcile every hour, suitable for resource that are not changed frequently
Medium (default) - commit check every 3 minutes, unconditional reconcile every 15 minutes
High - unconditional reconcile every 2 minutes for critical resources
Off - manual reconcile

By giving users option to select reconcile frequency options (high, medium, low and off) in channel configuration, the application or cluster admin can avoid unnecessary resource reconciliations and therefore prevent overload on subscription operator.

Also by adjusting the default resource reconciliation frequency to medium, we can improve initial application deployment performance significantly.

User should be able to override channel's reconcile frequency to OFF.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@
 
 # This repo is build in Travis-ci by default;
 # Override this variable in local env.
-TRAVIS_BUILD  ?= 1
+TRAVIS_BUILD  ?= 0
 
 # Image URL to use all building/pushing image targets;
 # Use your own docker registry and image name for dev/test by overridding the IMG and REGISTRY environment variable.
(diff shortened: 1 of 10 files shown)


xiangjingli (commented):
LGTM. Noticed the unit test code is not appended yet. The would be the next step.

PR #456, opened 2021-03-24
Fix appmgr pod crash

rokej:
appmgr pod crashed with 

```E0324 14:12:05.321204       1 manager.go:186] no matches for kind "Channel" in version "apps.open-cluster-management.io/v1"Manager exited non-zero```


mikeshng (approved):
/lgtm

PR #460, opened 2021-03-25
upgrade CRD version to v1, also keep the v1beta1 version for managed …

xiangjingli:
…cluster usage

https://github.com/open-cluster-management/backlog/issues/10642

The PR is to refresh app CRDs to v1, including:

1. apps.subscription CRD need to provide two versions - v1 used in ACM Hub.  V1 and V1beta1 are required in managed cluster. Helm chart will decide which version to be installed in managed cluster.
2.  refresh all other CRDs to v1, which are used in unit test and demo usage. 
3.  other fixes that caused the failure of unit test.


xiangjingli:
@mikeshng @rokej Please hold the PR, Just found the deployable V1 CRD have some other issues, that caused the failure of the unit test in deployable repo, though the unit test in application repo has passed. I will come up with new fix tomorrow.

mikeshng:
reopening in another PR because hung sonar scan

mikeshng:
reopening to see if hung prow jobs are fixed.

PR #465, opened 2021-04-01
Add cluster-name and cluster-sever labels to managed cluster secret

xiangjingli:

https://github.com/open-cluster-management/backlog/issues/10822


rokej (approved):
The code looks good but is a truncated server URL usable?

PR #480, opened 2021-04-23
AWS S3 object store support

xiangjingli:
https://github.com/open-cluster-management/backlog/issues/11490
https://github.com/open-cluster-management/backlog/issues/11301

The fix includes:

1. upgrade aws sdk to the latest version github.com/aws/aws-sdk-go-v2 v1.3.2 to support both minio and AWS S3 object buckets.

2. AWS s3 object bucket has to specify the region to the channel reference secret. Minio object bucket doesn't need this info, apply the default minio region info.

3. If the object buckect fails to connect, save the failure message to the appsub status.

4. apply non cached local/hub client ONLY for local/hub subscription status update as the latest updated local subscription status won't be fetched rigt away by cached client
As a result, subscription status fast conitinous update could easily fail when using cached client.
Since subscription status update does not usually happen, applying non cached client to it won't impact performance too much.

5. code enhancments to avoid controller crash when working with AWS S3 object store.


xiangjingli:
@mikeshng  

It seems the image built in the PR is not pushed to Quay. After the image built task is done, I can see the 2.3 and latest tag in the quay.io got refreshed, but I tried the new image in my env. It doesn't include my new code.

Here is what I grabbed from the image built log. I can see it does download the new version of aws lib and got the binary built, So the new image got built with the new binary. Just not sure where it goes :-)
```
...
STEP 5: RUN make -f Makefile.prow build
...
go: downloading github.com/aws/aws-sdk-go-v2 v1.3.2

real	0m23.972s
user	3m22.514s
sys	0m15.667s

real	0m2.459s
user	0m4.876s
sys	0m1.016s

....

Successfully pushed image-registry.openshift-image-registry.svc:5000/ci-op-y982xjfm/pipeline@sha256:a4b42df20477e6d0795730e778ccfac028b22ab840908d6e8bca6098bf1e5075
Push successful
``` 

Does that mean the image built in PR stage is only used for unit test. It won't be visible in quay until we merge the PR and get the publish task complete?  

Also I noticed tons of such warning in the image building. 
```
warning: file /usr/share/zoneinfo/zone1970.tab: remove failed: No such file or directory
warning: file /usr/share/zoneinfo/zone.tab: remove failed: No such file or directory
```
It seems the tzdata package is not installed correctly. I accessed one subscription container and found there is no `/usr/share/zoneinfo`

mikeshng:
responded to XJ's inquiry in slack.

ianzhang366:
After the PR merge, need to do some performance tests to see if the `noncachedclient` is called too often or not

PR #490, opened 2021-05-06
Remove orphan appsub

rokej:
https://github.com/open-cluster-management/backlog/issues/12033

When a managed cluster goes offline, placement decision changes to exclude the managed cluster and then the managed cluster comes back up, the subscription remains on the remote managed cluster even though the managed cluster has been removed from the placement decision.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@
 
 # This repo is build in Travis-ci by default;
 # Override this variable in local env.
-TRAVIS_BUILD  ?= 1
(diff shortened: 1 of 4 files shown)


xiangjingli (approved):

xiangjingli:
@rokej   FYI The [WIP] label blocked the auto merge.

rokej:
@xiangjingli Thanks. I added [WIP] because I found that it left deployed resources. Need to fix again.

xiangjingli (approved):

PR #493, opened 2021-05-14
add s3 object bucket subfolder support

mikeshng:

- allows subfolder object bucket subscription based on an annotation: api group `/bucket-path`
- perform an object bucket hub subscription "dryrun" to see which objects will be deployed when subscribed
- populate the topology annotation according according to the result of the dryrun. They are now categorized under "object" instead of previously "deployable"

see https://github.com/open-cluster-management/backlog/issues/12344 and https://github.com/open-cluster-management/backlog/issues/12510


xiangjingli (approved):
LGTM!

PR #503, opened 2021-06-01
fix manual sync

rokej:

https://github.com/open-cluster-management/backlog/issues/12965

image


mikeshng (approved):
/lgtm

PR #505, opened 2021-06-02
enable proxy for git clone

rokej:


xiangjingli:
@rokej  just a reminder. Please use `squad and merge` option  when you manually merge the PR. So all the commits in the PR will be squashed into single commit.

PR #511, opened 2021-06-09
do not default to amd64 if GOARCH is not set

dberg1:
If GOARCH environment variable is not set, use the local architecture instead of amd64.
Required for ppc64le support.


dberg1:
This is related to a change in downstream Dockerfile of multicluster-operators-subscription.
They changed the direct call to `go build` into a call to `make build`.

dberg1:
Notice that exact same change has already been made to other repos:
https://github.com/open-cluster-management/multicloud-operators-deployable/pull/82
https://github.com/open-cluster-management/multicloud-operators-placementrule/pull/85
https://github.com/open-cluster-management/multicloud-operators-channel/pull/163
https://github.com/open-cluster-management/multicloud-operators-application/pull/98

PR #512, opened 2021-06-09
add helmrepo fail status

rokej:

https://github.com/open-cluster-management/backlog/issues/13138

When the helm repo URL is incorrect, you will see hub subscription status like this with this change.

```
status:
  lastUpdateTime: "2021-06-09T14:06:31Z"
  phase: PropagationFailed
  reason: 'unable to retrieve the helm repo index https://artifacthub.io/packages/helm/bitnami/nginx: error unmarshaling JSON: while decoding JSON: json: cannot unmarshal string into Go value of type repo.IndexFile'
```


mikeshng (approved):
/lgtm
thanks Roke

PR #515, opened 2021-06-21
Add userid managed appsubs

philipwu08:
https://github.com/open-cluster-management/backlog/issues/9533

Signed-off-by: Philip Wu phwu@redhat.com


philipwu08:
@xiangjingli Found a couple of other things on Friday:
- appsub on the managed cluster (local-cluster) doesn't have the userid and group annotations. The user and group annotations are injected to the child appsub resource from the managed cluster, using the values in the annotation of the local appsub. So the appsub on the managed cluster needs theses annotations. I don't know if it creates any problems, or side-effects from this change.
- kubeResource doesn't contain "spec", so when a resource is created from it, it fails due to the spec validation. So instead, I unmarshal the resource to an unstructured object and inject the annotations that way.

xiangjingli (approved):
Just Peer reviewed with @philipwu08  LGTM

PR #518, opened 2021-06-22
Ansible jobs manual sync

rokej:
https://github.com/open-cluster-management/backlog/issues/13304

When the manual sync button is pressed on the app topology diagram or when the target commit or tag is changed, the ansible job hooks need to be run again.


ianzhang366:
code looks good, if we can have some test to cover it, it would be nicer

PR #524, opened 2021-06-25
support AWS S3 bucket pagination for listing over 1000 objects

xiangjingli:

https://github.com/open-cluster-management/backlog/issues/13642

AWS S3 enforces the 1000 object list per request. The fix is to apply pagination when listing objects to make sure all objects (> 1000) is able to be fetched.


mikeshng (approved):
/lgtm

PR #556, opened 2021-08-24
Secondary channel support

rokej:


mikeshng:
@rokej test failure might be because this branch is missing the https://github.com/open-cluster-management/multicloud-operators-subscription/pull/555 fix.

PR #572, opened 2021-09-13
fix synchronizer to apply only appsub resources

rokej:

Fix the synchronizer to search for specific resources with provided keys instead of looping though everything.


pkg/synchronizer/kubernetes/synchronizer.go
```diff
@@ -381,21 +381,17 @@ func isSpecialResource(gvr schema.GroupVersionResource) bool {
...
-			// if the current resource from the template map does not belong to ketSet,
-			// it belongs to another subscription. Skip.
-			klog.V(1).Infof("k: %v, does not belong to the order to be processed. skip", k)
-			continue
-		}
+	for resourceKey, okVal := range keySet {
+		tplunit := res.TemplateMap[resourceKey]
 
-		err := sync.applyTemplate(nri, res.Namespaced, k, tplunit, isSpecialResource(res.GroupVersionResource))
+		if okVal && tplunit != nil {
```
mikeshng, line 387: If `tplunit` is nil I feel like we should log a message here. Otherwise I am good with the change. Thanks.

PR #579, opened 2021-09-21
appsub admin namespace scoped deployment

rokej:

https://github.com/open-cluster-management/backlog/issues/16244

### DIFF
--- a/pkg/subscriber/git/git_subscriber.go
+++ b/pkg/subscriber/git/git_subscriber.go
@@ -137,6 +137,11 @@ func (ghs *Subscriber) SubscribeItem(subitem *appv1alpha1.SubscriberItem) error 
 		ghssubitem.clusterAdmin = true
 	}
 
+	if strings.EqualFold(subAnnotations[appv1alpha1.AnnotationCurrentNamespaceScoped], "true") {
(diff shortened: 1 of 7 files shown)


pkg/subscriber/git/git_subscriber.go
```diff
@@ -137,6 +137,11 @@ func (ghs *Subscriber) SubscribeItem(subitem *appv1alpha1.SubscriberItem) error
 		ghssubitem.clusterAdmin = true
 	}
 
+	if strings.EqualFold(subAnnotations[appv1alpha1.AnnotationCurrentNamespaceScoped], "true") {
```
xiangjingli, line 140: @rokej could you check if my understanding is correct? 

It seems user need to specify the new current-namespace-scoped annotation if they want to make sure all resources are deployed to the appsub namespace regardless of the namespace setting specified in the resource yaml.

if it is correct, we probably need a doc to let end user be aware of the new annotation since 2.4

rokej: @xiangjingli You are right. I am working on a doc for this user story https://github.com/open-cluster-management/backlog/issues/16244

PR #582, opened 2021-09-23
allow appsub admin to deploy policies

rokej:

Allow subscriptions created by a subscription admin to deploy/reconcile ACM policy resources.

Also fix subscription admin annotation verification so that when a non subscription admin tries to add `apps.open-cluster-management.io/cluster-admin: "true"` annotation to subscription, it gets removed.

### DIFF
--- a/pkg/utils/subscription.go
+++ b/pkg/utils/subscription.go
@@ -896,22 +896,19 @@ func AllowApplyTemplate(localClient client.Client, template *unstructured.Unstru
 // IsResourceAllowed checks if the resource is on application subscription's allow list. The allow list is used only
 // if the subscription is created by subscription-admin user.
 func IsResourceAllowed(resource unstructured.Unstructured, allowlist map[string]map[string]string, isAdmin bool) bool {
-	// Policy is not allowed by default
-	allowed := resource.GetAPIVersion() != "policy.open-cluster-management.io/v1"
-
 	// If subscription-admin, honor the allow list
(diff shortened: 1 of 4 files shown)


pkg/utils/subscription.go
```diff
@@ -896,22 +896,19 @@ func AllowApplyTemplate(localClient client.Client, template *unstructured.Unstru
 // IsResourceAllowed checks if the resource is on application subscription's allow list. The allow list is used only
 // if the subscription is created by subscription-admin user.
 func IsResourceAllowed(resource unstructured.Unstructured, allowlist map[string]map[string]string, isAdmin bool) bool {
-	// Policy is not allowed by default
-	allowed := resource.GetAPIVersion() != "policy.open-cluster-management.io/v1"
-
 	// If subscription-admin, honor the allow list
 	if isAdmin {
 		// If allow list is empty, all resources are allowed for deploy except the policy
```
xiangjingli, line 901: @rokej The comment here need to change accordingly, like 'If allow list is empty and user is subscription admin, all resources are allowed for deploy including the policy'  We can refresh the comment next time though

Please consider to cherry pick the PR to the new working upstream branch `AppSubStatusCRD2` when you have cycle.

rokej: @xiangjingli good catch. updated here https://github.com/open-cluster-management/multicloud-operators-subscription/pull/583

PR #592, opened 2021-10-06
Update policy generator version

dhaiducek:
Pre-emptive PR to address:
- https://github.com/open-cluster-management/backlog/issues/16925


build/Dockerfile
```diff
@@ -1,5 +1,5 @@
 FROM registry.ci.openshift.org/open-cluster-management/builder:go1.16-linux AS plugin-builder
-ENV POLICY_GENERATOR_TAG=v1.0.0
+ENV POLICY_GENERATOR_TAG=v1.0.1
```
mprahl, line 2: @dhaiducek could you please set this to `v1.1.0` instead?

I just cut a release for that. It had to be 1.1.0 since a new feature was added by the ZTP folks.
https://github.com/open-cluster-management/policy-generator-plugin/releases/tag/v1.1.0

PR #600, opened 2021-10-29
Fix resource cleanup for DR

rokej:

https://github.com/open-cluster-management/backlog/issues/17671


xiangjingli:
The fix looks good!  Just let me know when it can be merged to 2.4. I will set  the lgtm label to merge it

PR #608, opened 2021-12-07
Update the Policy Generator to v1.2.3

mprahl:
This is to include the following bug fix:
https://github.com/open-cluster-management/backlog/issues/18236


mikeshng (approved):
/lgtm

PR #619, opened 2021-12-15
Don't watch placement decision api until it is installed

xiangjingli:

This is to prevent the hub subscription pod crash when there is no placement decision CRD installed. The subscription community operator is installed before the  ACM cluster lifecycle component

### DIFF
--- a/cmd/manager/exec/manager.go
+++ b/cmd/manager/exec/manager.go
@@ -201,6 +201,9 @@ func RunManager() {
 
 	sig := signals.SetupSignalHandler()
 
+	klog.Info("Detecting ACM Placement Decision API...")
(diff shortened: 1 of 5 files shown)


xiangjingli:
@mikeshng  @philipwu08   The unit test in the PR failed. I noticed the same failure happened in other closed PR as well.  I run the test func TestClusterChange in placementrule_controller_test.go locally but the error doesn't happen.  The placementDecision status got the two decisions as expected. 

I will go ahead to merge the fix into midstream, and create the PR for upstream. Or the hub subscription pod gets crashed because there is no placementDecision API ready when only the appsub community operator is installed. That failed the appsub community operator CSV test task.

Could you take a look at the unit test errors when you have cycle?

PR #671, opened 2022-03-02
add mergeAndOwn option

rokej:

https://github.com/stolostron/backlog/issues/20167


rokej:
The unit-test failure is not related to this change.

PR #680, opened 2022-03-08
Add ansiblejob e2e test. 

mikeshng:
Its only responsible for verifying that ansiblejob prehook CR gets created and the CR name is in the subscription status.

sync upstream to midstream.


rokej:
@mikeshng so why is it an e2e test?

rokej (approved):

PR #701, opened 2022-03-21
Add appsub status reference to subcription CRD

philipwu08:
https://github.com/stolostron/backlog/issues/20331

### DIFF
--- a/pkg/controller/subscription/subscription_controller.go
+++ b/pkg/controller/subscription/subscription_controller.go
@@ -252,14 +252,12 @@ func (r *ReconcileSubscription) Reconcile(ctx context.Context, request reconcile
 			}
 
 			// Update AppstatusReference
-			if !r.standalone {
-				appsubStatusName := request.NamespacedName.Name
-				if strings.HasSuffix(appsubStatusName, "-local") {
-					appsubStatusName = appsubStatusName[:len(appsubStatusName)-6]
-				}
-
-				instance.Status.AppstatusReference = fmt.Sprintf("kubectl get appsubstatus -n %s %s", request.NamespacedName.Namespace, appsubStatusName)
+			appsubStatusName := request.NamespacedName.Name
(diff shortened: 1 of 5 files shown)


mikeshng:
It might be missing code gen changes? Same comment here https://github.com/open-cluster-management-io/multicloud-operators-subscription/pull/133#issuecomment-1074381152

pkg/controller/subscription/subscription_controller.go
```diff
@@ -250,6 +251,16 @@ func (r *ReconcileSubscription) Reconcile(ctx context.Context, request reconcile
 				klog.Errorf("doReconcile got error %v", reconcileErr)
 			}
 
+			// Update AppstatusReference
+			if !r.standalone {
```
xiangjingli, line 255: Does it mean the AppStatus reference is only visible for non standalone appsub?  It looks not necessary.

For standalone appsub, It looks the AppStatusreference is still valid. I just checked some standalone appsub e.g
```
% oc get appsub -n open-cluster-management          cluster-lifecycle-sub          
NAME                    STATUS       AGE     LOCAL PLACEMENT   TIME WINDOW
cluster-lifecycle-sub   Subscribed   5h11m   true      
```
And its appsubstatus is 
```
% oc get appsubstatus -n open-cluster-management          cluster-lifecycle-sub 
NAME                    AGE
cluster-lifecycle-sub   5h12m
```

xiangjingli (commented):
@philipwu08  BTW, after the PR is merged, I will create a new community CSV to bundle the new appsub CRD.  I will let you know when the new community CSV is merged. You can verify this using a new ACM upstream build

pkg/controller/subscription/subscription_controller.go
philipwu08: XJ, you're right with the standalone case. Let me update that.

PR #707, opened 2022-03-28
Use mch image manifest configmap for appsub addon image

philipwu08:
https://github.com/stolostron/backlog/issues/21013


mikeshng:
Same comment in https://github.com/open-cluster-management-io/multicloud-operators-subscription/pull/143#issuecomment-1081169733

PR #708, opened 2022-03-29
fix git clone error status

rokej:

Fix Git clone error status


mikeshng (approved):
/approve

/lgtm

PR #709, opened 2022-03-29
fix git clone error status

rokej:

Fix Git clone error status

### DIFF
--- a/pkg/utils/gitrepo.go
+++ b/pkg/utils/gitrepo.go
@@ -166,6 +166,8 @@ func getCertChain(certs string) tls.Certificate {
 	return certChain
 }
 
+// A subscription can have secondary channel to use when it cannot connect to the primary channel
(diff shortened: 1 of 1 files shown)


pkg/utils/gitrepo.go
```diff
@@ -170,7 +170,7 @@ func getConnectionOptions(cloneOptions *GitCloneOption, primary bool) (connectio
 	if !primary {
 		if cloneOptions.SecondaryConnectionOption == nil {
 			klog.Error("no secondary channel to try")
-			return nil, errors.New("no secondary channel to try")
+			return nil, nil
```
mikeshng, line 175: I traced thru the code and it seems like the primary getConnectionOptions error is now swallowed and not being returned in at least one logic path.

mikeshng (approved):
/approve

/lgtm

PR #729, opened 2022-05-03
always update token secret on reconcile

rokej:

https://github.com/stolostron/backlog/issues/22135

Whenever the controller reconciles appmgr SA token secret on a managed cluster, update the secret on the hub.


mikeshng (approved):
/lgtm

PR #730, opened 2022-05-04
Adding the Updated column for Subscription

jtyr:
This PR is adding another column into the `Subscription` CRD that will show the last time when the resource was updated. This little improvement helps people to debug problems a little bit easier.

It would be great if the `Subscription` resource would contain also the time of the last reconciliation that could be then added as yet another column. Unfortunately such information is not available in the `status` yet.


rokej:
Looks good to me.

PR #756, opened 2022-06-16
release-2.3: fix hub dry run for helm subscription not able to get secret and configmap properly

mikeshng:

for https://github.com/stolostron/backlog/issues/23541


mikeshng:
@xiangjingli please take another I had to make it fix to make sure the NS value weren't populated before overriding it with the channel NS value.

PR #786, opened 2022-07-08
Fix appv1 change

lennysgarage:
Signed-off-by: Jonathan Marcantonio <jmarcant@redhat.com>

Changed wrong revert from https://github.com/stolostron/multicloud-operators-subscription/pull/781#discussion_r916256235
line 818 should be err and line 830 should be nil as was originally 
Also changed missing removal of alpha1 from appv1 in helmrepo_test.go


mikeshng (approved):
/approve

/lgtm

PR #788, opened 2022-07-14
kube 1.24 service account token secret changes

rokej:

Kube 1.24 (OCP 4.11) changed the way service account token secret is mounted on the service account. The code is updated to handle this and also backward compatible.

https://github.com/stolostron/backlog/issues/24073

### DIFF
--- a/pkg/controller/spoketoken/spoke_token_controller.go
+++ b/pkg/controller/spoketoken/spoke_token_controller.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"encoding/json"
 	"net/url"
+	"strings"
 	"time"
 
 	ocinfrav1 "github.com/openshift/api/config/v1"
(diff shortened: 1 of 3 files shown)


pkg/controller/spoketoken/spoke_token_controller.go
```diff
@@ -244,27 +254,27 @@ func (r *ReconcileAgentToken) prepareAgentTokenSecret(token string) *corev1.Secr
...
+		klog.Error(err.Error())
+		return nil
+	}
 
-		// Get the service account token
-		if saSecret.Type == corev1.SecretTypeServiceAccountToken {
-			break
+	for _, secret := range secretList.Items {
+		// Get the application-manager service account token
+		if secret.Type == corev1.SecretTypeServiceAccountToken && strings.HasPrefix(secret.Name, "application-manager-token-") {
```
mikeshng, line 271: Since we are no longer relying on a direct linkage of ServiceAccount to Secret and instead just relying on the type and the name of the secret. It seems like if there is a secret with a prefix  `application-manager-token-` and the right type that anyone can create it will run into some kind of conflict?

@rokej I assume this issue is not unique to the app team. Is there another PR from another team that we can reference? Thanks.

rokej: Good point, I will add another annotation check `kubernetes.io/service-account.name: application-manager`.  Here is another PR foundation team did https://github.com/stolostron/managedcluster-import-controller/pull/177

rokej: @mikeshng I don't think you can create a service account type secret without actual service account.

rokej: That will be a kubernetes bug.

PR #804, opened 2022-07-25
Fix APIVersion in generateResourceList -author: chenz4027

mikeshng:


chenz4027 (approved):
lgtm, thanks for picking it up

PR #807, opened 2022-07-26
Add permission to delete placementdecisions for app SA

philipwu08:
https://github.com/stolostron/backlog/issues/24554


mikeshng:
@philipwu08 you shouldn't need to do this anymore and let the magic mirror sync for us.

/hold

xiangjingli:
@philipwu08  As Mike commented here
https://github.com/stolostron/multicloud-operators-subscription/pull/807#issuecomment-1195989155

You can create a new PR in the upstream repo. Once it is merged, the magic mirror will create a PR in the midstream repo for us.

PR #944, opened 2023-04-03
Remove cached client

lennysgarage:
This PR removes the cached client previously used in the application-manager controller.


xiangjingli:
The PR looks good. We will merge it once it is verified on e2e test

lennysgarage:
@xiangjingli made changes to the `getServiceAccountTokenSecret() `function and added the golang profiler webserver.

PR #1019, opened 2023-10-26
stop synchronizing resources list if any kustomization folder fails

xiangjingli:


mikeshng (commented):
/approve

I am ok with this change.
Leaving it to Roke for final approval.

mikeshng (approved):
/approve

/lgtm

We have to see if all the tests will pass.

PR #1020, opened 2023-10-26
stop synchronizing resources list if any kustomization folder fails (…

xiangjingli:
…#1019)


mikeshng (approved):
/approve

/lgtm

PR #1033, opened 2024-01-03
sync upstream

mikeshng:


build/Dockerfile
```diff
@@ -1,13 +1,13 @@
 FROM registry.ci.openshift.org/stolostron/builder:go1.20-linux AS plugin-builder
-ENV POLICY_GENERATOR_TAG=release-2.9
+ENV POLICY_GENERATOR_TAG=v1.12.4
```
dhaiducek, line 2: @mikeshng I'll revert this, but for your awareness, this shouldn't have been changed in `stolostron`.

dhaiducek: ...actually, this one is one us. We should have bumped it to `release-2.10` a while ago. I'll put that on our radar for next time.

PR #1236, opened 2025-03-31
[release-2.12]New annotation for hub sub bypass validation

mikeshng:

Should be same as https://github.com/stolostron/multicloud-operators-subscription/pull/1215


mikeshng:
backporting. scan not aware of e2e coverage.

PR #1299, opened 2025-05-07
use go env GOARCH when building go binaries on vairous platforms

xiangjingli:

https://issues.redhat.com/browse/ACM-20550#

The fix has been verified in latest downstream build


mikeshng (approved):
/lgtm

/approve

PR #1301, opened 2025-05-12
fix Konflux Enterprise contract violations

xiangjingli:
https://issues.redhat.com/browse/ACM-20662

### DIFF
--- a/.tekton/multicluster-operators-subscription-acm-214-pull-request.yaml
+++ b/.tekton/multicluster-operators-subscription-acm-214-pull-request.yaml
@@ -34,6 +34,12 @@ spec:
     value: build/Dockerfile.rhtap
   - name: path-context
     value: .
+  - name: build-source-image
+    value: "true"
+  - name: hermetic
+    value: "true"
+  - name: prefetch-input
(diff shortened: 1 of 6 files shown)


.tekton/multicluster-operators-subscription-acm-214-pull-request.yaml
```diff
@@ -117,6 +123,11 @@ spec:
         set of values is determined by the configuration of the multi-platform-controller.
       name: build-platforms
       type: array
+    - default: "true"
+      description: |
+        Check symlinks in the repo. If they're pointing outside of the repo, the build will fail.
+      name: enableSymlinkCheck
+      type: string
```
dhaiducek, line 130: This should be removed. The param to ignore symlinks is task-specific rather than pipeline-specific, so the one below this in `clone-repository` is all that should be needed.
```suggestion
```

dhaiducek:
In case it wasn't noticed already, `tzdata` is missing from the `rpms.lock.yaml` so isn't available to install in the hermetic environment:

```
Installed package tzdata-2025b-1.el9.noarch not available.
```

fxiang1:
@xiangjingli as Dale mentioned the `tzdata`(and possibly `git-core` & `openssh-clients`) package is missing from the rpm* files. You'll have to follow this doc https://konflux.pages.redhat.com/docs/users/building/prefetching-dependencies.html#rpm (VPN required) to add those to the rpm.in.yaml and then run `rpm-lockfile-prototype` to generate a new `rpm.lock.yaml` file.

PR #1336, opened 2025-07-03
Build the PolicyGenerator in RHEL8 for ArgoCD

dhaiducek:
This logic exists in CPaaS and needs to be ported to Konflux.


mikeshng:
Flakey test to be fix in another PR.

PR #1374, opened 2025-09-02
feat: adding argocd-agent to gitops addon requires the following install related manifests changes

mikeshng:
feat: adding argocd-agent to gitops addon requires the following install related manifests changes

see https://github.com/stolostron/multicloud-integrations/pull/386 for more details.


fxiang1:
/hold
In case @xiangjingli has any concerns.

PR #1562, opened 2026-05-19
git-path annotation path traversal allows cluster-admin mani…

xiangjingli:
…fest application

https://github.com/stolostron/multicloud-operators-subscription/security/advisories/GHSA-jfp9-h5p9-qrmx


xiangjingli:
since the case is still valid, the PR is merged to the main branch for ACM 5.0 and above as a normal bug fix (not CVE)

